Issue #24, opened 2026-07-30
Hosted AO v1: ao setup-vm part one, preflight and install


harshitsinghbhandari:
Batch 3, task 8 of the Hosted AO v1 build.

Spec: `docs/superpowers/specs/2026-07-29-hosted-ao-v1-accounts-and-machines.md` (section "`ao setup-vm`").

Scope:
- Ubuntu LTS with systemd and apt only. Anything else exits with the manual path.
- **Preflight, before mutating anything:** resolve the supplied hostname and confirm it points at this box public IP, confirm 80 and 443 are reachable from outside, confirm sudo. On failure, print exact remediation and change nothing. The cloud firewall rule is the one thing the script cannot fix, so it must detect and instruct.
- Installs the `ao` binary, tmux, git, gh, a systemd unit for the daemon, and a systemd unit for the gateway. It does **not** install harnesses.
- Idempotent and re-runnable.
- Ends with a summary of what is still missing (no harness configured, no git credentials) with the exact commands to fix each.

The gateway unit must set an absolute `DATA_DIR`. Per #18 the control plane now requires it, and the same discipline applies here: a working-directory-relative path under systemd is how keys and state silently relocate.

Out of scope: device binding and `machine.json` writing (task 10, batch 4), `ao whoami`, `ao vm setup-harness claude` (task 11).

Issue #23, opened 2026-07-30
Hosted AO v1: control plane device flow and machine registry


harshitsinghbhandari:
Batch 3, task 7 of the Hosted AO v1 build.

Spec: `docs/superpowers/specs/2026-07-29-hosted-ao-v1-accounts-and-machines.md` (sections "Control plane" and "The end-to-end flow").
Token contract: `controlplane/TOKEN_CONTRACT.md`, **as corrected in #18**. Read the current version.

Scope:
- RFC 8628 device authorization endpoints: device code issuance, the token polling endpoint with the server-supplied interval, expiry, and slow-down handling.
- The enter-code page and the approval confirmation page (`html/template`), reachable from the signed-in browser session added in #12.
- Machine registration on approval: insert into `machines`, return the machine id, account id, and public URL to the polling client.
- The authenticated list-machines API the desktop consumes in batch 4.

Critical correctness note, this is the exact trap #18 fixed a comment about: the JWT audience is **`machines.id`**, never `machines.hostname` or the public URL. Mint access tokens with `aud` = machine id via the existing `Issuer` from #13.

The `device_codes` and `machines` tables already exist in `0001_init.sql`, including the `status` CHECK constraint, the `user_code` UNIQUE constraint, `last_polled_at` for interval enforcement, and `revoked_at`. Add a migration only if a column is genuinely missing.

Out of scope: `ao setup-vm` (task 8, in parallel), the desktop machine list UI (task 12), machine revocation UI.

PR #27, opened 2026-07-30
feat(desktop): AO account login via system-browser PKCE

### LINKED ISSUE #25: Hosted AO v1: desktop AO login
Batch 3, task 9 of the Hosted AO v1 build.

Spec: `docs/superpowers/specs/2026-07-29-hosted-ao-v1-accounts-and-machines.md` (section "Desktop app").
Token contract: `controlplane/TOKEN_CONTRACT.md`, **as corrected in #18**. It now states the cookie name `ao_gw_token`, the `SameSite=None` requirement, the refresh token lifetime and rotate-on-use rule, and the corrected SSE transport line. Read the current version rather than inferring.

Scope:
- Login uses the **system browser** with PKCE and a loopback redirect (RFC 8252). Google blocks OAuth in embedded webviews, so an in-app `BrowserWindow` login is not available. Do not attempt one.
- The refresh token is encrypted with Electron `safeStorage` and written under `~/.ao/`. **Nothing** is written to an OS default app-data location; see the hard rule in `AGENTS.md` and the "App state lives under `~/.ao` only" section of `CLAUDE.md`. `frontend/src/main.ts` pins `userData` to `~/.ao/electron`; do not remove that override.
- Signed-in state and sign-out.
- `AO_CONTROL_URL` as the development hatch pointing at a locally-run control plane. It selects which control plane to trust; it can never skip authentication.

**Local use never requires an account.** Login is required only to reach a remote machine. Do not gate local mode behind sign-in.

`AO_REMOTE_URL` and `AO_REMOTE_TOKEN` keep working unchanged. They are removed only in the final batch, once the login path works end to end on a fresh VM. Do not delete them.

Out of scope: the machine list and machine switching (task 12, batch 4), authenticated remote transport (task 13, batch 5).

Follow `DESIGN.md`, starting with the "clone agent-orchestrator verbatim" banner at the top. Build new UI from shadcn primitives in `components/ui/*` where a component fits.

harshitsinghbhandari:
Closes #25. Item 9 of batch 3 in `docs/superpowers/specs/2026-07-29-hosted-ao-v1-accounts-and-machines.md`.

## What this adds

AO account sign-in in the desktop app, and nothing else. No machine list, no machine picker, no authenticated remote transport: those are tasks 12 and 13 in later batches.

**Login in the system browser, PKCE, loopback redirect (RFC 8252).** Google blocks OAuth in embedded webviews, so login goes through `shell.openExternal` and comes back to a listener on `127.0.0.1` on an ephemeral port. The verifier and `state` are 32 CSPRNG bytes each; the challenge is S256. The listener accepts exactly one callback, validates `state` before it reads the code, and shuts down immediately (including `closeAllConnections`, so the port is released rather than held by a browser keep-alive). A callback whose `state` does not match is rejected with a 400, not exchanged. Nothing logs the authorization code, the verifier, the refresh token, or an access token.

**Refresh token encrypted with `safeStorage`, under `~/.ao`.** Stored at `<~/.ao state dir>/ao-account.json`, mode 0600, atomic temp-file-plus-rename like `app-state.json`. The account id and email stay in the clear (identity, not a credential) so the signed-in row renders without a keychain unlock; the token itself is `safeStorage` ciphertext. When `safeStorage.isEncryptionAvailable()` is false, a real case on Linux without a Secret Service provider, the app refuses to sign in and says what to install, rather than writing a 90-day credential in plaintext. That check runs *before* a browser is opened.

**Signed-in state and a sign-out that discards the token.** Sign-out deletes the file, so a fresh controller cannot resurrect the credential. Covered by a test that asserts exactly that.

**`AO_CONTROL_URL` as the dev hatch.** It selects which control plane is trusted (default `https://ao.agentlab.in`; plain HTTP accepted only on the loopback, for `go run ./cmd/controlplane`). A malformed value is a hard error surfaced in the UI, not a silent fallback to production. When it is set to anything non-default the Settings row names it, so a dev hatch is never invisible. It cannot skip authentication and nothing added here can.

## Hard rules

- **Nothing to an OS default app-data location.** All new state resolves under the `~/.ao` state dir via the existing `runFilePath()` seam (so `AO_RUN_FILE` still redirects it). The `app.setPath("userData", ~/.ao/electron)` pin in `frontend/src/main.ts` is untouched.
- **Local use never requires an account.** The only new surface is a Settings section. Nothing gates startup, the local daemon, or the board, and there is no first-run sign-in wall. The section says so in copy: "Everything on this computer works without an account."
- **`AO_REMOTE_URL` and `AO_REMOTE_TOKEN` are untouched.** `shared/remote-daemon.ts`, `main/remote-daemon.ts`, and their wiring in `main.ts` are not modified by this PR.

## Control-plane contract

The desktop endpoints are being built in parallel under issue #23, so this builds against a written contract rather than waiting: `docs/desktop-login-contract.md` specifies `GET /oauth/desktop/authorize` and `POST /oauth/desktop/token`, the client id `ao-desktop`, and the token response `{ refresh_token, account: { id, email } }`. **If issue #23 lands different paths or field names, that doc and `frontend/src/main/ao-pkce.ts` change together.**

Login deliberately returns no access token: `aud` is a machine id per `controlplane/TOKEN_CONTRACT.md` and no machine is chosen at login time. The doc also records that refresh tokens rotate on use, and points task 13 at `writeStoredAccount` as the seam that must persist each replacement.

## UI

One new Settings section built from the existing `SettingsSection` / `SettingsRow` / `Button` primitives, in the same visual language as General and Updates, per the clone-agent-orchestrator-verbatim banner in `DESIGN.md`. Verified in the app with `ao preview http://localhost:5173/#/settings`.

Signed out: `AO account | Not signed in | [Sign in]`, plus the "works without an account" line.
Signed in: `AO account | dev@example.com | [Sign out]`, plus "Signing out discards the stored credential on this computer."

## Tests run

- `npx vitest run` (the CI command): **95 files, 1152 tests, all passing**, including 49 new ones across the PKCE helpers, the loopback listener, the on-disk store, the controller, the control-plane URL parser, the preload channels, and the settings section.
- `npm run typecheck` and `npm run typecheck:e2e`: clean. The e2e typecheck matters here because `satisfies AoBridge` in `e2e/support/fake-bridge.ts` is the preload-drift tripwire; both fake bridges and the jsdom setup bridge gained the `account` namespace.
- `CI=true npx playwright test --grep "@T0|@P0"` (renderer-smoke): **20 passed**. SET-001 now asserts the Account section renders and reads "Not signed in", since local use must not require an account.

Notable test cases: the verifier sent to the token endpoint is the pre-image of the challenge the browser was sent; the refresh token never appears in plaintext on disk and the file is 0600; nothing at all is written when encryption is unavailable; a forged `state` is rejected and no token is stored; an unreachable control plane reports plainly and points at local mode; a second Sign in click joins the attempt already in flight instead of opening a second tab.

## Not in scope

No `controlplane/` or `backend/` changes (issues #23 and #24 own those), so no HTTP surface change: `openapi.yaml` and `frontend/src/api/schema.ts` are untouched, as intended.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/desktop-login-contract.md
+++ b/docs/desktop-login-contract.md
@@ -1,0 +1,95 @@
+# Desktop login contract
+
+The two control-plane endpoints the desktop app's sign-in talks to. Written by the
+desktop side (issue #25) while the control plane's own routes were being built in
+parallel (issue #23), so the app builds against this contract rather than waiting.
+If the control plane lands different paths or field names, this file and
+`frontend/src/main/ao-pkce.ts` change together.
+
+Token shapes, lifetimes, and the rotate-on-use rule live in
+[`controlplane/TOKEN_CONTRACT.md`](../controlplane/TOKEN_CONTRACT.md); this file is
+only the desktop login exchange.
+
+## Why this shape
+
+The desktop app is a **public client**: it ships no secret. Google blocks OAuth in
+embedded webviews, so login runs in the user's real browser and comes back through
+an RFC 8252 loopback redirect. PKCE (S256) plus a CSPRNG `state` is the whole
+defence against another local process stealing the authorization code.
+
+`AO_CONTROL_URL` selects which control plane the app trusts (default
+`https://ao.agentlab.in`; plain HTTP is accepted only on the loopback, for a
+locally-run control plane). It can never skip authentication, and there is no flag
+that does.
+
+## `GET /oauth/desktop/authorize`
+
+Opened in the system browser. Query parameters:
+
+| Parameter | Value |
+| --- | --- |
+| `response_type` | `code` |
+| `client_id` | `ao-desktop` |
+| `redirect_uri` | `http://127.0.0.1:<ephemeral>/callback` |
+| `code_challenge` | base64url(SHA-256(verifier)) |
+| `code_challenge_method` | `S256` |
+| `state` | 32 CSPRNG bytes, base64url |
+
+The control plane signs the user in with Google (its existing `/auth/google/login`
+session), then redirects to `redirect_uri` with `code` and the unmodified `state`,
+or with `error` plus optional `error_description`.
+
+The redirect URI is a loopback address on an ephemeral port, per RFC 8252 section
+7.3, so the control plane must accept any port for `127.0.0.1` on this client. The
+desktop listener accepts exactly one callback, validates `state` before reading the
+code, and shuts down immediately.
+
+## `POST /oauth/desktop/token`
+
+`Content-Type: application/x-www-form-urlencoded`:
+
+```
+grant_type=authorization_code
+client_id=ao-desktop
+code=<from the callback>
+code_verifier=<the PKCE verifier>
+redirect_uri=<the same loopback URI>
+```
+
+Success, `200`, JSON:
+
+```json
+{
+  "refresh_token": "<opaque, high entropy>",
+  "account": { "id": "<accounts.id>", "email": "<google email>" }
+}
+```
+
+Failure returns a non-2xx with the OAuth error envelope `{"error": "...",
+"error_description": "..."}`; `invalid_grant` covers a reused code, a mismatched
+verifier, and a mismatched `redirect_uri`.
+
+**No access token is issued here.** An access token's `aud` is a machine id
+(`controlplane/TOKEN_CONTRACT.md`) and no machine is selected at login time, so
+access tokens come from a later refresh exchange scoped to the chosen machine. That
+exchange, along with attaching credentials to REST, SSE, and `/mux`, is task 13 in
+batch 5 and is deliberately not implemented here.
+
+## What the desktop stores
+
+`~/.ao/ao-account.json`, mode 0600 (the `~/.ao` state dir, `AO_RUN_FILE`-relative,
+never an OS default app-data location):
+
+```json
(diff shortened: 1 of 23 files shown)


harshitsinghbhandari:
Updated for the orchestrator's two-audience decision (control-plane audience for control-plane API calls, machine audience per machine).

Nothing here had to change behaviorally, and nothing was stubbed:

- The refresh token already goes to exactly one place, `POST /oauth/desktop/token`. It is never a Bearer credential on a resource route, never in a URL, never logged.
- No access token of either audience is fetched or cached in this task, so there is nothing beyond the stored refresh token for sign-out to discard today. `signOut` deletes the file, which is all of it.
- Token storage was never shaped around one universal token: `ao-account.json` holds only the refresh token plus identity.

What did change is the contract doc, which asserted the single-audience assumption and would have misled the next reader: `docs/desktop-login-contract.md` now has a "Two token audiences, deliberately not interchangeable" section, defers to `controlplane/TOKEN_CONTRACT.md` once hosted-ao-17 adds its own section there, and records that **when task 13 caches a control-plane-audience access token, sign-out must clear that cache too**. The same note sits on `signOut` in `frontend/src/main/ao-account.ts` so it is visible at the call site, not only in docs.

PR #97, opened 2026-08-18
docs: record the pair-mode live test findings and the pivot to mobile's model

harshitsinghbhandari:
Supersedes the transport half of the pair-by-IP spec (#86).

A live end-to-end run against a real Debian box proved the gateway, passcode
auth, per-source lockout, and the control-route allowlist all work. It also
found five defects no unit test caught, including one that leaves **every fresh
box dead on arrival** and one that made certificate pinning reject every
connection after pairing.

**Decision: drop TLS and pinning. Reuse the daemon's LAN listener, the one the
phone already talks to.** Plaintext HTTP, passcode auth, no certificates. That
is ADR 0001's posture, and it is the trade already accepted for mobile.

The trade, stated plainly: plaintext means the passcode and everything else
travel in the clear, so this is LAN-only and a public cloud VM is no longer a
valid test target.

This doc is the handoff: what shipped, what the live test proved, what it found,
and the delete/keep/build/rewrite list. ADR 0003 and the AGENTS.md carve-out
both need superseding.

### DIFF
--- a/docs/plans/2026-08-18-pair-mode-status-and-pivot.md
+++ b/docs/plans/2026-08-18-pair-mode-status-and-pivot.md
@@ -1,0 +1,151 @@
+# Pair mode: what shipped, what the live test found, and the pivot
+
+Written 2026-08-18, at the point where the TLS/pinning design was abandoned in
+favour of Connect Mobile's model. Supersedes the transport half of
+`2026-08-16-pair-by-ip-headless-boxes.md`. Read this before touching pair mode.
+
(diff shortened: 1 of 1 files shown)


harshitsinghbhandari:
Closing as obsolete. Current develop and accepted ADR 0004 retain self-signed TLS, SHA-256 fingerprint pinning, and support for cloud VMs. This older plaintext, LAN-only pivot contradicts the current architecture and must not land.

PR #98, opened 2026-08-19
chore: merge upstream/main (230 commits, base 5319744)

harshitsinghbhandari:
# Merge upstream/main into develop (230 commits)

## Sizing

- Merge base: 53197448f (2026-08-07), verified healthy (ancestor of both sides, proportionate divergence).
- Files touched by both sides (conflict surface): 87. Only ours: 151. Only theirs: 1056.
- Rollback point: tag `pre-upstream-merge-20260819`.

## How the 87 shared files were resolved

- 5 regenerate-not-merge (openapi.yaml, schema.ts, lockfiles): took a side, then regenerated; regeneration was byte-identical (MD5-verified), so the committed content is generator output.
- 18 take-theirs mechanical (.gitignore, README translations, go.mod/sum, i18n JSONs; fork i18n keys restored in a follow-up commit).
- 13 take-theirs with the small fork hunk reapplied (state-root constants, observer credential stripping, auto-updater guard, doc paragraphs).
- 51 hosting seam files resolved hunk-by-hunk. Every must-survive item verified by an independent review with file:line evidence: state root under ~/.ao/hosted, GET /api/v1/doctor, machines.* IPC, runtimeFetch/SSE gateway rebasing, session-level TLS pinning, cloneUrl wire, credential stripping, Hosted AO packaging identity, pair-mode telemetry meta, CSP connect-src https:/wss:.

## Decisions made in this merge

- **Upstream AO Cloud is pinned permanently hidden** in Hosted AO builds (single shared constant, reversible; decision recorded in CONTEXT.md and the 2026-08-19 onboarding spec). Upstream's cloud code rides along dormant so future merges stay cheap. Scope of the pin: the renderer sign-in entry, the `ao-app://` OS protocol claim (removed from the bundle; upstream uses it for AO Cloud deep links, and claiming it would hijack stock agent-orchestrator's sign-in callback on shared machines), the protocol registration, and the deep-link routing. Dormant but present: the WorkOS code, `cloud:*` IPC handlers, and `cloud-auth.bin` under `~/.ao/hosted`.
- **Two upstream behaviors deliberately adopted over fork ones**: `WORKSPACE_CHILD_DEFAULT_BRANCH_UNKNOWN` dropped (upstream removed it deliberately), and `DefaultBranch` now resolves via upstream's `gitdefault` instead of the fork's checked-out-branch heuristic.
- **Clone-by-URL UX now forks on machine locality** (`isRemoteDaemonBaseUrl`, https-only): remote machines lose the local destination picker and clone into the machine's managed repos dir; local machines keep upstream's picker flow unchanged. Phase 1 warning carried in the ledger: a plaintext-LAN transport would silently break this classifier.
- **Remote clone-by-URL got a real fix, not a casualty**: upstream's new clone dialog uses the desktop's local folder picker, which cannot work against a remote machine. When the active machine is remote, the dialog now submits the Git URL over the fork's surviving `cloneUrl` wire on POST /api/v1/projects (51-line conditional path; local flow byte-identical to upstream).
- **Doctor cluster was ported, not merged**: upstream's new GitLab check moved into our internal/doctor with PublicMessage redaction so usernames never cross the gateway.
- **Upstream's `resolveGitOriginURL` was deliberately dropped** in favor of our `gitremote.OriginURL`: upstream does not strip credentials from recorded origins; we do.
- New migrations from upstream land as-is; zero ordinal collisions (84 migrations verified from empty). Note: upstream now occupies ordinal 0100, so the spec's "hosted-only migrations start at >= 0100" convention gets a new floor (>= 0200) before Phase 1 writes one.

## New guard rails added

- `test(vmgateway)`: pins that `/api/v1/mobile/devices` (upstream's new device roster) is blocked at the gateway.
- `test(desktop)`: state-root guard test, catches any future `os.homedir()` + ".ao" join that skips the hosted segment; this exact class arrived twice via clean auto-merges in this merge.
- Vitest pinning that the Cloud sign-in entry cannot render.

## Verification

- npm run lint, frontend typecheck (0 errors), full vitest suite (221 files, 2845 passed, 0 failed), go test ./internal/httpd/... ./internal/storage/... (719 passed).
- The four playbook invariant greps clean.
- Packaged build on Node 22: dmg + zip produced (built via an `ELECTRON_MIRROR` workaround for a local-network TLS failure reaching objects.githubusercontent.com; not a repo change).
- Live smoke against a real machine: still outstanding at PR-open time; called out explicitly rather than implied by the green gates. Will be run before landing (or recorded as an intentional gap).

## Known deferrals (tracked in the SDD ledger for follow-up issues)

- Gateway `blockedAPIPrefixes` lacks `/api/v1/browser` while the daemon LAN listener blocks it; pre-existing on develop, comment claims a mirror falsely. Separate hardening issue; not changed mid-merge.
- `.clone-progress-bar` in styles.css currently unused (Phase 3 cleanup).
- `ao-control-token.test.ts` real-clock 20ms deadline can flake under suite load; wants fake timers.
- Cloud pin's unit test asserts the gate function, not component non-rendering (the render gate is a single line; trace verified manually).


harshitsinghbhandari:
History note: this PR was squash-merged as 30da98de6 because the repo only allowed squash merges at the time. That flattened the 230 upstream commits and would have destroyed the merge base for every future upstream sync, so develop was rewound and re-merged as the true merge commit 3a5b1327b (parents 6d7c0f170 + 1dcc04e7c, tree identical to the squash). Repo settings now allow merge commits; upstream sync PRs must always merge with a merge commit, never squash.

PR #117, opened 2026-08-20
fix(desktop): stop pairing probe from wedging a paired box offline

### LINKED ISSUE #114: Paired machine reads as unreachable until the app is restarted, despite a correct pin and a healthy box
## Bug

After completing the Add machine pairing flow, the newly paired box shows the `daemon_unreachable` banner ("<name> is not reachable, and has never connected to AO") for the rest of the app session. The stored record is correct and the box is healthy the whole time. Quitting and relaunching the app fixes it immediately, with no change to the box, the network, or the stored data.

**Analyzed against:** `0b570fcd8822b8522ac171818dbd28b861157aa4` (develop, post-[#113](https://github.com/agentlab-in/hosted-ao/pull/113))
**Version:** desktop 0.13.1, macOS arm64, locally signed build
**Confidence:** Medium. Evidence rules out the obvious causes and the restart-fixes-it behaviour is consistent, but the caching mechanism itself is not directly proven.

## Reproduction

1. Provision a box for pair mode (`ao pair`), note the pairing string.
2. In the desktop app, Add machine, paste the string, compare and accept the fingerprint.
3. Observe the machine is added and the record persisted correctly.
4. Observe the machine reads as unreachable, indefinitely, for the rest of the session.
5. Quit and relaunch the app. The machine is reachable within one probe cycle.

Observed live against a pair-provisioned box at `20.197.63.75:443`.

## Evidence

The persisted record was already correct at the time the banner was showing, in `~/.ao/hosted/ao-paired-machines.json`: right address and port, both address hints, and a `pinnedFingerprint` exactly equal to the certificate served by the box, verified independently with `openssl s_client`. `lastSeen` was `null`, so no probe had ever completed.

The box was reachable from the same machine throughout: TCP 443 open, TLS serving, and `GET /api/v1/doctor` answering `401` without credentials.

That 401 is the key discriminator. `paired-machines.ts:384` treats any HTTP response, including a non-ok one, as `online`:

```ts
if (!response.ok) return { reachability: "online", harness: HARNESS_UNKNOWN, answered: true };
```

So an `offline` verdict can only come from the `catch` at `paired-machines.ts:386-389`, a transport or TLS failure, never an auth failure. This also rules out a stale or rotated passcode, which would surface as 401 and therefore online.

After a relaunch with no other change, `lastSeen` was written and the machine came up online, confirming the credential and the pin were both valid the entire time.

## Root Cause

The gateway certificate is self-signed, `CN=ao-pair-gateway`, with `DNS:ao-pair-gateway` as its only SAN and no IP SAN. Connecting by IP therefore fails Chromium's default verification, and the pin is enforced instead by the verify proc installed on `session.defaultSession` at `src/main.ts:2345`.

By design, that proc **rejects** the certificate during pairing. `src/main/paired-machine-cert.ts:97-108` has no accept path for a host with nothing pinned yet: it captures the presented fingerprint via `onPresented` so the UI can show it, then returns `CERT_VERIFY_REJECT`. This is deliberate and correct, it is what makes trust-on-first-use a user decision rather than a transport decision.

The hypothesis is that this rejection populates Chromium's per-session certificate-error cache for that host. Once the user accepts and `add()` pins the fingerprint, the proc would return `CERT_VERIFY_ACCEPT`, but the cached failure means it is not consulted again for the life of the session. Restarting clears the cache and the first probe succeeds.

Note that the registry side is **not** at fault and was ruled out by reading: `add()` rebuilds the host to pin map at `paired-machines.ts:464` and explicitly clears the stale reachability verdict at `paired-machines.ts:474-475`, returning `unknown`. So this is not plain in-memory staleness.

## Confirming the mechanism

Remove and re-pair a machine in a running app. If it stays unreachable until relaunch, the bug reproduces. To confirm the cache specifically, log the verify proc on every invocation and check whether it is called at all for that host after the pin is stored. Silence there confirms Chromium is short-circuiting.

## Fix

Needs a way to invalidate the cached certificate error for the host after a pin is accepted, before the first post-pairing probe. Options worth evaluating:

- `session.clearHostResolverCache()` plus closing idle sockets for the host after `add()` resolves.
- `session.clearAuthCache()` / `clearData` scoped to the origin, if it covers cached cert errors.
- Avoid the poisoning entirely: run the fingerprint-capture probe on a throwaway `session.fromPartition()` so the rejection never touches the session that carries real traffic. This looks like the cleanest fix and keeps the security property intact.

## Impact

First-run pairing appears to fail every time. The user completes the flow correctly, accepts a fingerprint that matches, and is told the box is unreachable, with a hint that points at the wrong problem entirely (see below). Recovering requires quitting the app, which nothing in the UI suggests. This is the first thing a new user does with pair mode.

Secondary: the hint rendered under the `daemon_unreachable` banner is `daemon.hint.conflict`, "Stop the conflicting daemon, then restart the desktop app." That is a local-daemon-conflict message and is misleading for an unreachable remote machine. Worth splitting the hint per code.

## Related

- [#82](https://github.com/agentlab-in/hosted-ao/issues/82) renderer keeps querying the gateway with no bearer, silent 401 loop until relaunch. Same "wedged until relaunch" family, different layer.
- [#81](https://github.com/agentlab-in/hosted-ao/issues/81) design work on unifying local and self-hosted machines.

harshitsinghbhandari:
## Summary

Fixes #114: a freshly paired machine reads as unreachable (`daemon_unreachable`) for the rest of the app session, even though the persisted record and the pin are both correct and the box is healthy. Quitting and relaunching fixes it immediately with no other change, which is what the issue's evidence points at -- an in-memory cache that isn't invalidated when a pin is accepted mid-session.

**Root cause**, confirmed against the issue's analysis: `probeFingerprint()` (the trust-on-first-use capture connection the "Add machine" flow uses to show the presented fingerprint for comparison) is *always* denied at the TLS layer -- `paired-machine-cert.ts` has no accept path for a host with nothing pinned yet. That denial ran on `session.defaultSession`, the same session every other authenticated request to the box uses. Chromium caches a per-host certificate-error verdict at the network-service level, below where `session.setCertificateVerifyProc` runs, for the life of the session. Once the probe's rejection landed in that cache, every later connection to the host short-circuited straight to a transport failure -- the verify proc was never consulted again, so pinning the correct fingerprint in `add()` had no effect on already-cached hosts. `probeReachability()`'s doctor probe treats a transport failure as "offline" (correctly -- it's the only path that legitimately means offline), so the machine stayed stuck there until a relaunch cleared the cache.

**Fix**: route `probeFingerprint`'s capture connection through a session other than `session.defaultSession` (`Session.fetch()`, not `net.fetch()` -- the latter always issues from the default session). The probe's rejection now dies with that session instead of poisoning the one real traffic uses; `session.defaultSession` sees the paired host for the first time only *after* `add()` has pinned the correct fingerprint, so it verifies and accepts on the very first connection. This was one of the three options the issue raised and the one it flagged as cleanest, since it avoids the poisoning entirely rather than trying to invalidate a cache after the fact.

`paired-machines.ts` gains an optional `probeNetFetch` dependency (defaults to `netFetch` for existing callers/tests) used only by `probeFingerprint`; `netFetch` itself is now used only for calls against an already-pinned host (`refresh`'s doctor probe). `main.ts`'s `pairProbeNetFetch` keys the throwaway session by target hostname rather than minting a fresh one per call, since `session.fromPartition` has no destroy API (electron/electron#27142, #28566) and a fresh partition every probe would leak one session per candidate address raced, unbounded for the life of the app; keying by host bounds that to one throwaway session per distinct box ever probed.

**Known residual gap**, filed as #121: because the probe session is now reused per host, a *second* probe of a still-unpinned host in the same app session (the realistic case: abandon a pairing attempt after comparing the fingerprint, retry) can hit its own cached rejection and misreport "no certificate could be retrieved." Re-pairing an already-registered box is unaffected. Left as a follow-up rather than folding a second, more involved mitigation (bumping the session key only on a capture miss) into this PR.

**Out of scope**: the issue's secondary note about `daemon.hint.conflict` being a misleading hint for an unreachable *remote* machine (it's written for a local-daemon-conflict) is a separate UI-copy fix unrelated to this root cause; leaving it for a follow-up to keep this PR focused on the reachability bug.

## Test plan

- [x] Added two regression tests to `paired-machines.test.ts` that model Chromium's per-session certificate-error cache: one reproduces #114 by sharing a session between the probe and real traffic (probe's rejection wedges `refresh()` offline after a correct `add()`), the other proves the fix by giving the probe its own session (`refresh()` comes back online with the pin correctly enforced on first contact).
- [x] `npx vitest run src/main` -- 527 passed, 1 skipped (all main-process tests, including every `paired-machine*` suite).
- [x] `cd frontend && npm run typecheck` -- no new errors from this change (pre-existing, unrelated `../packages/product-ui` errors reproduce identically on `develop`).

### DIFF
--- a/frontend/src/main.ts
+++ b/frontend/src/main.ts
@@ -1857,14 +1857,18 @@ function aoMachines(): ReturnType<typeof createAoMachinesController> {
  * tries, and again on every manual retry -- for as long as the app stays
  * open. Keying by host instead bounds that to one throwaway session per
  * distinct box ever probed, which is what actually needs isolating from
- * `session.defaultSession`. The cost: a probe of a host that was already
- * probed and rejected once in this session -- still unpinned, the box's cert
- * genuinely rotated between the two attempts -- can hit its own cached
- * rejection and fail to (re)capture the new fingerprint until the app
- * restarts. That is a narrower recurrence of the same class of bug, scoped
- * to an already-rare event (a paired box re-minting its long-lived
- * certificate between two probes of the same unpinned host, in one app
- * session) rather than to the every-single-pairing case #114 was.
+ * `session.defaultSession`.
+ *
+ * The cost: a probe of an unpinned host is always denied, by design (same as
+ * above), so reusing that host's session for a second probe can hit its own
+ * cached rejection and fail to capture anything -- not only when the box's
+ * certificate has actually changed between the two attempts. The realistic
+ * trigger is a retry after an abandoned or failed pairing attempt (compare
+ * the fingerprint, close the dialog without pinning, try again), which can
+ * come back with a misleading "no certificate could be retrieved" until the
+ * app restarts. Re-pairing an already-registered box is unaffected: a pinned
+ * host verifies and accepts on `session.defaultSession`, which this probe
+ * session never touches. Tracked as a follow-up: #121.
  */
 function pairProbeNetFetch(input: RequestInfo | URL, init?: RequestInit): Promise<Response> {
 	const hostname = new URL(String(input)).hostname;
--- a/frontend/src/main.ts
+++ b/frontend/src/main.ts
@@ -1830,15 +1830,13 @@ function aoMachines(): ReturnType<typeof createAoMachinesController> {
 // app.whenReady below, is what actually enforces the pin; this controller only
 // holds the data it reads from.
 
-let pairProbeSessionSeq = 0;
-
 /**
- * Fetch for `probeFingerprint`'s capture connection, bound to a fresh,
- * in-memory session every call rather than `session.defaultSession`.
- * `ses.fetch()`, not `net.fetch()`: the latter always issues from the default
- * session (that is the whole reason it exists as a convenience wrapper), so
- * routing through another session requires calling `fetch` on the `Session`
- * instance itself.
+ * Fetch for `probeFingerprint`'s capture connection, bound to a session other
+ * than `session.defaultSession`, keyed by the target host rather than a
+ * fresh partition every call. `ses.fetch()`, not `net.fetch()`: the latter
+ * always issues from the default session (that is the whole reason it exists
+ * as a convenience wrapper), so routing through another session requires
+ * calling `fetch` on the `Session` instance itself.
  *
  * That connection is always denied at the TLS layer (paired-machine-cert.ts
  * has no accept path for a host with nothing pinned yet), and Chromium caches
@@ -1849,12 +1847,28 @@ let pairProbeSessionSeq = 0;
  * rejection: the very next authenticated request to the newly paired host
  * would still be short-circuited to a failure before the verify proc is even
  * consulted, reading as unreachable until the app restarts and the cache is
- * gone (#114). A throwaway partition, discarded after the one probe it was
- * created for, means the poisoned verdict dies with it instead of outliving
- * the pairing flow.
+ * gone (#114).
+ *
+ * The partition is keyed by hostname, not minted fresh per call, because
+ * `session.fromPartition` has no destroy API: Electron holds every partition
+ * it has ever created in an internal map for the life of the process
+ * (electron/electron#27142, #28566), so a fresh name every call would leak
+ * one Session per probe -- one per candidate address `racePairAddresses`
+ * tries, and again on every manual retry -- for as long as the app stays
+ * open. Keying by host instead bounds that to one throwaway session per
+ * distinct box ever probed, which is what actually needs isolating from
+ * `session.defaultSession`. The cost: a probe of a host that was already
+ * probed and rejected once in this session -- still unpinned, the box's cert
+ * genuinely rotated between the two attempts -- can hit its own cached
+ * rejection and fail to (re)capture the new fingerprint until the app
+ * restarts. That is a narrower recurrence of the same class of bug, scoped
+ * to an already-rare event (a paired box re-minting its long-lived
+ * certificate between two probes of the same unpinned host, in one app
+ * session) rather than to the every-single-pairing case #114 was.
  */
 function pairProbeNetFetch(input: RequestInfo | URL, init?: RequestInit): Promise<Response> {
-	const probeSession = session.fromPartition(`pair-probe-${pairProbeSessionSeq++}`);
+	const hostname = new URL(String(input)).hostname;
+	const probeSession = session.fromPartition(`pair-probe-${hostname}`);
 	probeSession.setCertificateVerifyProc(pairedMachines().verifyCertificate);
 	return probeSession.fetch(String(input), init);
 }
--- a/frontend/src/main.ts
+++ b/frontend/src/main.ts
@@ -1829,6 +1829,36 @@ function aoMachines(): ReturnType<typeof createAoMachinesController> {
 // plane. session.defaultSession's certificate-verify proc, installed in
 // app.whenReady below, is what actually enforces the pin; this controller only
 // holds the data it reads from.
+
+let pairProbeSessionSeq = 0;
+
+/**
+ * Fetch for `probeFingerprint`'s capture connection, bound to a fresh,
+ * in-memory session every call rather than `session.defaultSession`.
+ * `ses.fetch()`, not `net.fetch()`: the latter always issues from the default
+ * session (that is the whole reason it exists as a convenience wrapper), so
+ * routing through another session requires calling `fetch` on the `Session`
+ * instance itself.
+ *
+ * That connection is always denied at the TLS layer (paired-machine-cert.ts
+ * has no accept path for a host with nothing pinned yet), and Chromium caches
+ * a per-host certificate-error verdict at the network-service level, below
+ * where the verify proc runs, for the life of the session that made the
+ * connection. If the capture probe shared `session.defaultSession` with real
+ * traffic, pinning the fingerprint in `add()` would not undo that cached
+ * rejection: the very next authenticated request to the newly paired host
+ * would still be short-circuited to a failure before the verify proc is even
+ * consulted, reading as unreachable until the app restarts and the cache is
+ * gone (#114). A throwaway partition, discarded after the one probe it was
+ * created for, means the poisoned verdict dies with it instead of outliving
+ * the pairing flow.
+ */
+function pairProbeNetFetch(input: RequestInfo | URL, init?: RequestInit): Promise<Response> {
+	const probeSession = session.fromPartition(`pair-probe-${pairProbeSessionSeq++}`);
+	probeSession.setCertificateVerifyProc(pairedMachines().verifyCertificate);
+	return probeSession.fetch(String(input), init);
+}
+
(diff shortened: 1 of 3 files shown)


harshitsinghbhandari (commented):
This PR fixes the root cause of #114 (a paired machine reading as unreachable for the rest of the app session after a fresh pairing) by giving the pairing probe its own throwaway Electron session instead of sharing session.defaultSession with real traffic. The fix is well targeted, and the two new regression tests in paired-machines.test.ts model the Chromium per-session certificate-error cache convincingly -- one reproduces #114 by sharing a session between probe and real traffic, the other proves the fix by giving the probe its own session -- both exercised through the same public API (probeFingerprint/add/refresh) rather than mocking internals.

One concern worth addressing before merge: pairProbeNetFetch in main.ts mints a brand new session.fromPartition(`pair-probe-${pairProbeSessionSeq++}`) on every single probe call, and Electron never releases sessions created via fromPartition for the life of the process -- there is no destroy API (see electron/electron#27142 and #28566, a known and still-open limitation). Since probeFingerprint is called once per candidate address by racePairAddresses (frontend/src/shared/pair-race.ts) on every pasted pairing string, and again on every manual retry, this permanently accumulates one leaked Session (and its network context) per attempt for as long as the app stays open -- exactly the kind of long-running desktop process where this compounds, in the exact code path that already had one production bug from session-level caching behavior. Left inline on the spot in main.ts.

Everything else here looks solid: the doc comments accurately describe the Chromium caching mechanism, the probeNetFetch/netFetch split is scoped correctly to the two call sites (capture probe vs. already-pinned doctor probe), and defaulting probeNetFetch to netFetch when omitted keeps every existing test in the file passing unchanged.

Requesting changes for the session leak above; nothing else blocks this.

frontend/src/main.ts
```diff
@@ -1829,6 +1829,36 @@ function aoMachines(): ReturnType<typeof createAoMachinesController> {
...
+ * traffic, pinning the fingerprint in `add()` would not undo that cached
+ * rejection: the very next authenticated request to the newly paired host
+ * would still be short-circuited to a failure before the verify proc is even
+ * consulted, reading as unreachable until the app restarts and the cache is
+ * gone (#114). A throwaway partition, discarded after the one probe it was
+ * created for, means the poisoned verdict dies with it instead of outliving
+ * the pairing flow.
+ */
+function pairProbeNetFetch(input: RequestInfo | URL, init?: RequestInit): Promise<Response> {
+	const probeSession = session.fromPartition(`pair-probe-${pairProbeSessionSeq++}`);
```
harshitsinghbhandari: session.fromPartition with a unique partition string per call means this session (and its underlying network context) is never released: Electron holds fromPartition sessions in an internal map for the life of the process, with no public API to destroy one (electron/electron#27142, #28566). racePairAddresses (frontend/src/shared/pair-race.ts) fires one probeFingerprint call per candidate address in a pasted pairing string, and every manual retry from the add-machine dialog adds another, so this leaks one Session per probe attempt for as long as the app stays open.

Consider keying the partition by the target host instead of a global incrementing counter (e.g. `pair-probe-${new URL(String(input)).hostname}`), so repeated attempts against the same box reuse one session rather than minting a new one every time. The only cost is that a certificate rotated between two probes of the same still-pinned-nothing host would hit its own stale cached rejection -- a much narrower version of the failure this PR already fixes.

harshitsinghbhandari: Fixed: keyed the throwaway probe session by hostname (`pair-probe-${hostname}`) instead of an incrementing counter, so repeated probes of the same box reuse one session and the leak is bounded to one throwaway session per distinct host ever probed, not one per call. Documented the tradeoff you flagged (a probe of a host whose cert rotated between two probes of the same still-unpinned host can hit its own cached rejection) directly in the updated doc comment on `pairProbeNetFetch` in main.ts -- accepted as a much narrower recurrence of the class of bug this PR fixes. Pushed in 358e9b826; `npx vitest run src/main` (527 passed) and `tsc --noEmit` both clean.

harshitsinghbhandari:
Reviewed. Root cause and fix both look right, and routing the capture connection off `session.defaultSession` is the correct shape: it avoids the poisoning rather than trying to invalidate a cache after the fact.

Good catch on the partition leak. `session.fromPartition` having no destroy API is exactly the trap a naive "throwaway session per probe" walks into, and keying by hostname bounds it to one session per distinct box. That is better than what the issue suggested.

## One finding: the residual case is broader than the doc comment says

The comment scopes the remaining wedge to a box "re-minting its long-lived certificate between two probes of the same unpinned host", which reads as very rare. I do not think certificate rotation is actually required for it.

`probeFingerprint` (`paired-machines.ts:541-562`) depends entirely on the verify proc firing to populate `presented` via `onPresented`. If a cached rejection short-circuits below the proc, nothing is captured and the function falls through to:

> "No certificate could be retrieved from that address. Check the address, port, and that the box is reachable."

So **any second probe of a still-unpinned host within one app session fails**, regardless of whether the certificate changed. The first probe always populates the cache, because an unpinned host is always denied by design.

The realistic trigger is not rotation, it is a first attempt that never completed: paste the pairing string, reach the fingerprint compare step, close or cancel the dialog, then try again. That user is now wedged until they restart the app, and the error tells them to check the address, the port, and whether the box is reachable, when in fact the box is fine. That misdirection is the part I would most want to avoid shipping, since it sends someone debugging their network for a client-side cache.

Worth being explicit that this does **not** affect re-pairing an already-registered box: a pinned host verifies and accepts normally, so that path is unaffected. It is specifically the never-completed-pairing retry.

## Suggestion

Bumping the partition key only when a probe fails to capture a fingerprint would fix the retry case while keeping the leak bounded by the number of failed attempts, which is small and self-limiting. Something along the lines of a per-host attempt counter incremented on a capture miss, so a retry gets a clean session but a healthy pairing still reuses one.

If you would rather keep this PR focused, that is reasonable. In that case please just correct the doc comment so it describes the real trigger (a retry after an abandoned or failed attempt) rather than certificate rotation, and open a follow-up. An accurate comment matters here because the next person to hit this will be reading exactly that paragraph.

## Minor

The PR body still ends with the "Generated with Claude Code" footer. This repo does not carry AI attribution. The commits themselves are clean, so this is a body edit only, no rewrite needed.

harshitsinghbhandari:
Thanks -- taking the "keep this PR focused" option, as offered.

- Fixed the doc comment on `pairProbeNetFetch` (frontend/src/main.ts) to describe the actual trigger: any second probe of a still-unpinned host in one app session can hit its own cached rejection and fail to capture, not only when the certificate rotates. Spelled out the realistic case (retry after an abandoned pairing attempt) and noted re-pairing an already-registered box is unaffected (pushed in 7488a2790).
- Filed #121 to track a proper fix, with your suggested direction (bump the per-host session key only on a capture miss, so the leak stays bounded by failed attempts rather than reused indefinitely) recorded there for whoever picks it up.
- Removed the "Generated with Claude Code" footer from the PR body -- noted, this repo doesn't carry AI attribution.

`npx vitest run src/main` (527 passed) and `tsc --noEmit` both still clean after the comment fix.

PR #134, opened 2026-08-28
fix: block system install routes at VM gateway

harshitsinghbhandari:
## Summary
- add `/api/v1/system/install` to the VM gateway shared blocked API-prefix policy
- return the existing 404 envelope before auth/proxying in both hosted-token and pair-passcode modes
- cover base, target, and nested paths across GET, POST, and DELETE while asserting zero daemon calls

## Security boundary
`/api/v1/system/install/{target}` mutates the host and is intentionally loopback-only. The LAN listener already denies this route family; this change closes the equivalent remotely exposed VM/pair gateway boundary without changing the loopback daemon route, public bind/auth semantics, or allowed REST/SSE/mux traffic. Existing middleware ordering is preserved.

hao Batch 1 depends on this prerequisite boundary fix after Hosted AO account removal.

## Tests
- `cd backend && go test ./internal/vmgateway`
- `env -u AO_SESSION_ID -u AO_PROJECT_ID -u AO_TMUX_BINARY -u AO_TMUX_SOCKET_NAME npm run lint`

### DIFF
--- a/backend/internal/vmgateway/passcode_test.go
+++ b/backend/internal/vmgateway/passcode_test.go
@@ -288,6 +288,28 @@ func TestPairGateway_ControlRoutesStayBlocked_EvenWithValidPasscode(t *testing.T
 	}
 }
 
+func TestPairGateway_SystemInstallRouteFamily_NeverReachesDaemon(t *testing.T) {
+	tg := newTestPairGateway(t)
+	for _, tc := range []struct {
+		method string
+		path   string
+	}{
+		{http.MethodGet, "/api/v1/system/install"},
+		{http.MethodPost, "/api/v1/system/install/gh"},
+		{http.MethodDelete, "/api/v1/system/install/gh/history"},
+	} {
+		t.Run(tc.method+" "+tc.path, func(t *testing.T) {
+			rec := tg.request(tc.method, tc.path, tg.passcode, "203.0.113.5:1")
+			if rec.Code != http.StatusNotFound {
+				t.Fatalf("status = %d, want 404 (even with a valid passcode)", rec.Code)
+			}
+		})
+	}
+	if len(tg.daemonCalls) != 0 {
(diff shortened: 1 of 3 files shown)


harshitsinghbhandari (commented):
## Security review

Verdict: no actionable issues found. This is approval-worthy, but GitHub does not permit the PR author to submit an approving review.

- `/api/v1/system/install` is added to the shared `blockedAPIPrefixes` policy used by both hosted-token and pair-passcode handlers.
- `hasPathPrefix` enforces an exact segment boundary, so the base path and every nested path are denied while siblings such as `/api/v1/system/installer` remain proxyable.
- The deny policy is method-independent. The added hosted and pair tests exercise GET, POST, and DELETE across the base, target, and deeper nested shapes and assert that the real reverse-proxy boundary records zero daemon calls.
- `denyByDefault` remains outside both authentication middlewares, so blocked requests return the existing 404 envelope without invoking JWT/JWKS or passcode/lockout processing. CORS remains intentionally outermost; its existing blocked-preflight test verifies that an allowed-origin preflight for a blocked family is also 404 and never proxied.
- The change is additive to the blocked-prefix list and does not alter the allow rules or transport/auth handling for allowed REST, SSE, notification/workspace streams, or `/mux`.
- The policy matches the LAN listener's existing `/api/v1/system/install` route-family block and the documented rule that host-mutating installer routes remain loopback-only.

Boundary verification: PASS. The changed policy producer (`blockedAPIPrefixes`) is consumed by `isProxyablePath` through `denyByDefault`; both gateway modes are exercised through their real handler chains into a real test reverse-proxy target, with the terminal invariant (404 and zero upstream calls) asserted. I applied the aggressive boundary review in-memory because this review was explicitly read-only and therefore did not generate report files.

I also confirmed PR head `75ecb870` is based directly on current `develop` `26962813`, and all reported checks are green.

PR #140, opened 2026-08-29
Add read-only hao status and doctor commands

### LINKED ISSUE #139: Add read-only hao status and doctor commands
Implement HAO Batch 3: a bounded, read-only machine observation layer with `hao status` and `hao doctor`.

Scope and acceptance criteria are defined by the HAO Batch 3 observation prompt. The implementation must preserve loopback confinement, bounded probes, redaction, stable DTOs/check IDs, strict exit semantics, AO doctor truth reuse, and zero mutation.

harshitsinghbhandari:
## Summary
- add versioned read-only `hao status` observations with strict exit semantics
- add independent `hao doctor` host checks and aggregate AO daemon doctor truth
- confine daemon HTTP probes to discovered loopback with timeouts, redirect rejection, and response bounds
- add injected observer coverage for deterministic machine/network tests

## Tests
- `cd backend && go test ./internal/haocli`

## Review focus
Please focus on observation truthfulness, loopback confinement/SSRF, timeout/body bounds, recursive redaction, strict exit semantics, AO doctor reuse, platform/service behavior, and artifact packaging.

Closes #139

### DIFF
--- a/backend/internal/haocli/doctor.go
+++ b/backend/internal/haocli/doctor.go
@@ -2,8 +2,7 @@ package haocli
 
 import (
 	"context"
-	"crypto/sha256"
-	"encoding/hex"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"os"
@@ -136,25 +135,29 @@ func buildDoctor(ctx context.Context, deps Deps, explicit string) (DoctorReport,
 		add(DiagnosticCheck{ID: "ao.doctor", Severity: "warning", Status: "unknown", Evidence: "AO doctor endpoint is unavailable", Remediation: "restore daemon reachability and rerun hao doctor"})
 	} else {
 		normalizedNames := make(map[string]map[string]struct{}, len(d.Doctor.Checks))
+		rawNameCounts := make(map[string]int, len(d.Doctor.Checks))
 		for _, check := range d.Doctor.Checks {
 			normalized := stableID(check.Name)
 			if normalizedNames[normalized] == nil {
 				normalizedNames[normalized] = map[string]struct{}{}
 			}
 			normalizedNames[normalized][check.Name] = struct{}{}
+			rawNameCounts[check.Name]++
 		}
-		seenNames := make(map[string]bool, len(d.Doctor.Checks))
+		duplicateReported := false
 		for _, check := range d.Doctor.Checks {
-			normalized := stableID(check.Name)
-			id := "ao.doctor." + normalized
-			if len(normalizedNames[normalized]) > 1 {
-				id += "-" + shortHash(check.Name)
-			}
-			if seenNames[check.Name] {
-				add(DiagnosticCheck{ID: "ao.doctor.invalid-duplicate-" + shortHash(check.Name), Severity: "error", Status: "error", Evidence: "AO doctor returned a duplicate check name"})
+			if rawNameCounts[check.Name] > 1 {
+				if !duplicateReported {
+					add(DiagnosticCheck{ID: "ao.projection.duplicate-doctor-names", Severity: "error", Status: "error", Evidence: "AO doctor returned duplicate check names"})
+					duplicateReported = true
+				}
 				continue
 			}
-			seenNames[check.Name] = true
+			normalized := stableID(check.Name)
+			id := "ao.doctor." + normalized
+			if len(normalizedNames[normalized]) > 1 || normalized == "" {
+				id += "-" + losslessID(check.Name)
+			}
 			status, severity := "unknown", "warning"
 			if check.Level == "WARN" {
 				status, severity = "warn", "warning"
@@ -279,9 +282,12 @@ func stableID(value string) string {
 	return strings.Trim(b.String(), "-")
 }
 
-func shortHash(value string) string {
-	sum := sha256.Sum256([]byte(value))
-	return hex.EncodeToString(sum[:4])
+func losslessID(value string) string {
+	encoded := base64.RawURLEncoding.EncodeToString([]byte(value))
+	if encoded == "" {
+		return "empty"
+	}
+	return encoded
 }
 
 func writeDoctorReport(cmd *cobra.Command, report DoctorReport) error {
--- a/backend/internal/haocli/doctor.go
+++ b/backend/internal/haocli/doctor.go
@@ -111,7 +111,7 @@ func buildDoctor(ctx context.Context, deps Deps, explicit string) (DoctorReport,
 	}
 
 	add(anyToolCheck(deps, "host.package-manager", []string{"apt-get", "dnf", "yum", "brew", "apk"}, "install a supported package manager or manage dependencies manually"))
-	add(anyToolCheck(deps, "host.service-manager", []string{"systemctl", "launchctl"}, "manage services manually on this platform"))
+	add(serviceManagerCheck(deps, goos))
 	add(diagnosticFromObservation(toolObservation(ctx, deps, "tool.git", "git", true, "--version")))
 	github := configString(object, "workflow", "profile") == "github"
 	add(diagnosticFromObservation(toolObservation(ctx, deps, "tool.gh", "gh", github, "--version")))
@@ -182,6 +182,21 @@ func anyToolCheck(deps Deps, id string, names []string, remediation string) Diag
 		}
 	}
 	return DiagnosticCheck{ID: id, Severity: "warning", Status: "unsupported", Evidence: "no supported implementation found", Remediation: remediation}
+}
+func serviceManagerCheck(deps Deps, goos string) DiagnosticCheck {
+	var manager string
+	switch goos {
+	case "linux":
+		manager = "systemctl"
+	case "darwin":
+		manager = "launchctl"
+	default:
+		return DiagnosticCheck{ID: "host.service-manager", Severity: "warning", Status: "unsupported", Evidence: "no supported service manager is defined for " + goos, Remediation: "manage services manually on this platform"}
+	}
+	if path, err := deps.Observer.LookPath(manager); err == nil {
+		return passCheck("host.service-manager", filepath.Base(path)+" is available")
+	}
+	return DiagnosticCheck{ID: "host.service-manager", Severity: "warning", Status: "unsupported", Evidence: manager + " is unavailable", Remediation: "manage services manually on this platform"}
 }
 func diagnosticFromObservation(o Observation) DiagnosticCheck {
 	status, severity := mapObservationStatus(o.Status), "info"
--- a/backend/internal/haocli/doctor.go
+++ b/backend/internal/haocli/doctor.go
@@ -113,8 +113,18 @@ func buildDoctor(ctx context.Context, deps Deps, explicit string) (DoctorReport,
 	add(diagnosticFromObservation(toolObservation(ctx, deps, "tool.git", "git", true, "--version")))
 	github := configString(object, "workflow", "profile") == "github"
 	add(diagnosticFromObservation(toolObservation(ctx, deps, "tool.gh", "gh", github, "--version")))
+	if github {
+		add(exitOnlyProbe(ctx, deps, "tool.gh.auth", "gh", []string{"auth", "status"}, "run gh auth login directly as the target user"))
+	} else {
+		add(DiagnosticCheck{ID: "tool.gh.auth", Severity: "info", Status: "disabled", Evidence: "GitHub workflow profile is not selected"})
+	}
 	harness := configString(object, "harness", "id")
 	add(diagnosticFromObservation(toolObservation(ctx, deps, "harness."+harness, harnessBinary(harness), true, "--version")))
+	if harness == "claude-code" {
+		add(exitOnlyProbe(ctx, deps, "harness.claude-code.auth", "claude", []string{"auth", "status", "--json"}, "run claude login directly as the target user"))
+	} else {
+		add(DiagnosticCheck{ID: "harness." + stableID(harness) + ".auth", Severity: "warning", Status: "unsupported", Evidence: "no safe authentication probe is defined for the selected harness"})
+	}
 
 	d := observeDaemon(ctx, deps.Observer, discoveryPath(stateRoot), deps.Timeout)
 	if d.State == "healthy" {
@@ -173,6 +183,22 @@ func anyToolCheck(deps Deps, id string, names []string, remediation string) Diag
 		}
 	}
 	return DiagnosticCheck{ID: id, Severity: "warning", Status: "unsupported", Evidence: "no supported implementation found", Remediation: remediation}
+}
+func exitOnlyProbe(ctx context.Context, deps Deps, id, binary string, args []string, remediation string) DiagnosticCheck {
+	path, err := deps.Observer.LookPath(binary)
+	if err != nil {
+		return DiagnosticCheck{ID: id, Severity: "error", Status: "fail", Evidence: binary + " is unavailable", Remediation: remediation}
+	}
+	probeCtx, cancel := boundedContext(ctx, deps.Timeout)
+	defer cancel()
+	_, err = deps.Observer.Run(probeCtx, path, args...)
+	if err == nil {
+		return passCheck(id, "authentication probe succeeded")
+	}
+	if probeCtx.Err() != nil || errors.Is(err, context.DeadlineExceeded) {
+		return DiagnosticCheck{ID: id, Severity: "warning", Status: "unknown", Evidence: "authentication probe timed out", Remediation: remediation}
+	}
+	return DiagnosticCheck{ID: id, Severity: "error", Status: "fail", Evidence: "authentication probe reported not authenticated", Remediation: remediation}
 }
 func diagnosticFromObservation(o Observation) DiagnosticCheck {
 	status, severity := mapObservationStatus(o.Status), "info"
--- a/backend/internal/haocli/doctor.go
+++ b/backend/internal/haocli/doctor.go
@@ -1,0 +1,239 @@
+package haocli
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+type DiagnosticCheck struct {
+	ID          string `json:"id"`
+	Severity    string `json:"severity"`
+	Status      string `json:"status"`
+	Evidence    string `json:"evidence"`
+	Remediation string `json:"remediation,omitempty"`
+}
+
+type DoctorReport struct {
+	SchemaVersion     int               `json:"schemaVersion"`
+	Machine           string            `json:"machine"`
+	Mode              string            `json:"mode"`
+	Checks            []DiagnosticCheck `json:"checks"`
+	Failures          int               `json:"failures"`
+	ExecutionFailures int               `json:"executionFailures"`
+}
+
+func newDoctorCommand(deps Deps, opts *options) *cobra.Command {
+	return &cobra.Command{Use: "doctor", Short: "Run read-only machine diagnostics", Args: noArgs,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			report, err := buildDoctor(cmd.Context(), deps, opts.configPath)
+			if err != nil {
+				return err
+			}
+			if opts.json {
+				err = writeJSON(cmd.OutOrStdout(), haocontractRedact(report))
+			} else {
+				err = writeDoctorReport(cmd, report)
+			}
+			if err != nil {
+				return err
+			}
+			if report.Failures > 0 || report.ExecutionFailures > 0 {
+				return commandError{Code: "doctor_failed", ExitStatus: 1, Silent: true}
+			}
+			return nil
+		}}
+}
+
+func buildDoctor(ctx context.Context, deps Deps, explicit string) (DoctorReport, error) {
+	path, object, err := loadConfig(deps, explicit)
+	if err != nil {
+		return DoctorReport{}, err
+	}
+	stateRoot, err := deps.StateDir()
+	if err != nil {
+		return DoctorReport{}, operationalError("resolve state root", err)
+	}
+	report := DoctorReport{SchemaVersion: observationSchemaVersion, Machine: configString(object, "machine", "name"), Mode: configString(object, "mode")}
+	add := func(check DiagnosticCheck) {
+		report.Checks = append(report.Checks, check)
+		if check.Status == "fail" {
+			report.Failures++
+		}
+		if check.Status == "error" {
+			report.ExecutionFailures++
+		}
+	}
+
+	goos, arch := deps.Observer.Platform()
+	if supportedPlatform(goos, arch) {
+		add(passCheck("host.platform", goos+"/"+arch+" is supported"))
+	} else {
+		add(DiagnosticCheck{ID: "host.platform", Severity: "warning", Status: "unsupported", Evidence: goos + "/" + arch + " is not a supported managed platform", Remediation: "use a supported Linux or macOS amd64/arm64 host"})
+	}
+	checkPath := func(id, target string, private bool) {
+		info, statErr := deps.Observer.Stat(target)
+		if statErr != nil {
+			status := "error"
+			severity := "error"
+			if errors.Is(statErr, os.ErrNotExist) {
+				status, severity = "fail", "error"
+			}
+			add(DiagnosticCheck{ID: id, Severity: severity, Status: status, Evidence: "could not inspect " + target, Remediation: "ensure the path exists and is readable by the target user"})
+			return
+		}
+		if !info.Owner {
+			add(DiagnosticCheck{ID: id, Severity: "error", Status: "fail", Evidence: "path is not owned by the current target user", Remediation: "correct ownership outside hao, then retry"})
+			return
+		}
+		if private && info.Mode.Perm()&0o077 != 0 {
+			add(DiagnosticCheck{ID: id, Severity: "error", Status: "fail", Evidence: fmt.Sprintf("permissions %04o are broader than owner-only", info.Mode.Perm()), Remediation: "restrict the path to the target user"})
+			return
+		}
+		add(passCheck(id, "ownership and permissions are acceptable"))
+	}
+	checkPath("state.root", stateRoot, false)
+	checkPath("config.file", path, true)
+	if available, diskErr := deps.Observer.Disk(stateRoot); diskErr != nil {
+		add(errorCheck("host.disk", "disk availability probe failed"))
+	} else if available < 512*1024*1024 {
+		add(DiagnosticCheck{ID: "host.disk", Severity: "error", Status: "fail", Evidence: fmt.Sprintf("%d bytes available", available), Remediation: "free at least 512 MiB on the state volume"})
+	} else {
+		add(passCheck("host.disk", fmt.Sprintf("%d bytes available", available)))
+	}
+
+	add(anyToolCheck(deps, "host.package-manager", []string{"apt-get", "dnf", "yum", "brew", "apk"}, "install a supported package manager or manage dependencies manually"))
+	add(anyToolCheck(deps, "host.service-manager", []string{"systemctl", "launchctl"}, "manage services manually on this platform"))
+	add(diagnosticFromObservation(toolObservation(ctx, deps, "tool.git", "git", true, "--version")))
+	github := configString(object, "workflow", "profile") == "github"
+	add(diagnosticFromObservation(toolObservation(ctx, deps, "tool.gh", "gh", github, "--version")))
+	harness := configString(object, "harness", "id")
+	add(diagnosticFromObservation(toolObservation(ctx, deps, "harness."+harness, harnessBinary(harness), true, "--version")))
+
+	d := observeDaemon(ctx, deps.Observer, discoveryPath(stateRoot), deps.Timeout)
+	if d.State == "healthy" {
+		add(passCheck("ao.daemon", "loopback health and readiness succeeded"))
+	} else if d.State == "unavailable" && !configBool(object, "service", "enabled") {
+		add(DiagnosticCheck{ID: "ao.daemon", Severity: "info", Status: "disabled", Evidence: "daemon is absent and service is intentionally disabled"})
+	} else {
+		add(DiagnosticCheck{ID: "ao.daemon", Severity: "error", Status: mapObservationStatus(d.State), Evidence: d.Evidence, Remediation: "inspect the AO daemon service and logs"})
+	}
+	if d.Doctor == nil {
+		add(DiagnosticCheck{ID: "ao.doctor", Severity: "warning", Status: "unknown", Evidence: "AO doctor endpoint is unavailable", Remediation: "restore daemon reachability and rerun hao doctor"})
+	} else {
+		for _, check := range d.Doctor.Checks {
+			status, severity := "pass", "info"
+			if check.Level == "WARN" {
+				status, severity = "warn", "warning"
+			}
+			if check.Level == "FAIL" {
+				status, severity = "fail", "error"
+			}
+			add(DiagnosticCheck{ID: "ao.doctor." + stableID(check.Name), Severity: severity, Status: status, Evidence: check.Message, Remediation: check.Remediation})
+		}
+	}
+	if report.Mode == "pair" {
+		port := configInt(object, "pair", "listenPort")
+		probeCtx, cancel := boundedContext(ctx, deps.Timeout)
+		available, portErr := deps.Observer.PortAvailable(probeCtx, "127.0.0.1", port)
+		cancel()
+		if portErr != nil {
+			add(errorCheck("gateway.port", "port availability probe failed"))
+		} else if !available {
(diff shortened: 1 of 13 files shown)


backend/internal/haocli/observe.go
```diff
@@ -0,0 +1,247 @@
...
+func (systemObserver) Stat(path string) (FileObservation, error)      { return statFile(path) }
+func (systemObserver) Disk(path string) (uint64, error)               { return diskAvailable(path) }
+func (systemObserver) LookPath(name string) (string, error)           { return exec.LookPath(name) }
+func (systemObserver) ReadRunFile(path string) (*runfile.Info, error) { return runfile.Read(path) }
+func (systemObserver) ProcessAlive(pid int) bool                      { return processalive.Alive(pid) }
+
+func (systemObserver) Run(ctx context.Context, name string, args ...string) (string, error) {
+	cmd := exec.CommandContext(ctx, name, args...)
+	cmd.Stdin = nil
+	var out bytes.Buffer
```
harshitsinghbhandari: `bytes.Buffer` is unbounded while the child runs; truncating after `cmd.Run()` does not bound memory. A noisy or hostile `--version` executable can allocate until OOM within the timeout. Capture through a fixed-cap writer (while continuing to drain/discard excess so the child cannot block) and add a real subprocess test that emits well beyond the limit.

backend/internal/haocli/status.go
```diff
@@ -0,0 +1,177 @@
...
+func buildStatus(ctx context.Context, deps Deps, explicit string) (StatusReport, error) {
+	path, object, err := loadConfig(deps, explicit)
+	if err != nil {
+		return StatusReport{}, err
+	}
+	stateRoot, err := deps.StateDir()
+	if err != nil {
+		return StatusReport{}, operationalError("resolve state root", err)
+	}
+	report := StatusReport{SchemaVersion: observationSchemaVersion, Machine: configString(object, "machine", "name"), Mode: configString(object, "mode"), ConfigPath: path, ConfigVersion: 1, HAOVersion: Version, DesiredAOVersion: configString(object, "components", "aoVersion")}
```
harshitsinghbhandari, line 76: This reconstructs `<stateRoot>/running.json` and therefore ignores a supported `AO_RUN_FILE` override whose basename is not literally `running.json`. `ResolveStateRoot` intentionally returns only the override's directory, losing the filename. Resolve the exact run-file path through the shared config boundary and cover `AO_RUN_FILE=/tmp/custom-daemon.json` end to end; `doctor.go:119` has the same bug.

backend/internal/haocli/status.go
```diff
@@ -0,0 +1,177 @@
...
+	d := observeDaemon(ctx, deps.Observer, discoveryPath(stateRoot), deps.Timeout)
+	desiredService := configBool(object, "service", "enabled")
```
harshitsinghbhandari, line 78: When the service is intentionally disabled and discovery is absent, status publishes `unavailable` with `desired:false`, while doctor correctly publishes `disabled`. Normalize this case to `disabled`; consumers should not have to infer intentional disablement from a second field, and the two commands must not disagree about the same state.

backend/internal/haocli/observe.go
```diff
@@ -0,0 +1,247 @@
...
+			return getErr
+		}
+		if len(body) > maxHTTPBody {
+			return errors.New("response exceeds size limit")
+		}
+		if err := json.Unmarshal(body, target); err != nil {
+			return fmt.Errorf("malformed response")
+		}
+		return nil
+	}
```
harshitsinghbhandari, line 188: All health transport failures inherit the pre-set `unhealthy` state, including deadline expiry/DNS- or socket-level inability to observe. A timeout proves neither bad health nor a dead daemon; classify context deadline/cancellation as `unknown` (and preserve `unavailable` for proven absence) just as `toolObservation` already does. Add an assertion on `ao.daemon`, not merely a substring check that can be satisfied by `ao.readiness`.

backend/internal/haocli/doctor.go
```diff
@@ -0,0 +1,265 @@
...
+	} else if d.State == "unavailable" && !configBool(object, "service", "enabled") {
+		add(DiagnosticCheck{ID: "ao.daemon", Severity: "info", Status: "disabled", Evidence: "daemon is absent and service is intentionally disabled"})
+	} else {
+		add(DiagnosticCheck{ID: "ao.daemon", Severity: "error", Status: mapObservationStatus(d.State), Evidence: d.Evidence, Remediation: "inspect the AO daemon service and logs"})
+	}
+	if d.Doctor == nil {
+		add(DiagnosticCheck{ID: "ao.doctor", Severity: "warning", Status: "unknown", Evidence: "AO doctor endpoint is unavailable", Remediation: "restore daemon reachability and rerun hao doctor"})
+	} else {
+		for _, check := range d.Doctor.Checks {
+			status, severity := "pass", "info"
```
harshitsinghbhandari: Unknown/malformed daemon `level` values fall through as `pass`. Because this endpoint is explicitly unversioned and daemon-defined, a newer or corrupt level must fail closed as `unknown`/execution error, never become healthy. Switch exhaustively on PASS/WARN/FAIL and test an unrecognized value through the real daemon DTO decoder.

backend/internal/haocli/doctor.go
```diff
@@ -0,0 +1,265 @@
...
+			if check.Level == "WARN" {
+				status, severity = "warn", "warning"
+			}
+			if check.Level == "FAIL" {
+				status, severity = "fail", "error"
+			}
+			add(DiagnosticCheck{ID: "ao.doctor." + stableID(check.Name), Severity: severity, Status: status, Evidence: check.Message, Remediation: check.Remediation})
```
harshitsinghbhandari: Two boundary failures meet here. First, daemon-controlled message/remediation strings are emitted after a redactor that misses `Authorization: Bearer ...` and JSON-embedded secrets; an adversarial report leaks both in human and JSON output. Second, lossy slugging makes distinct names such as `runtime capability` and `runtime/capability` publish the same check ID. Strengthen recursive string redaction for bearer/pairing/PEM/JSON forms and preserve a collision-free daemon ID (or detect/reject duplicates), with adversarial tests.

backend/internal/haocli/doctor.go
```diff
@@ -0,0 +1,265 @@
...
+	if available, diskErr := deps.Observer.Disk(stateRoot); diskErr != nil {
+		add(errorCheck("host.disk", "disk availability probe failed"))
+	} else if available < 512*1024*1024 {
+		add(DiagnosticCheck{ID: "host.disk", Severity: "error", Status: "fail", Evidence: fmt.Sprintf("%d bytes available", available), Remediation: "free at least 512 MiB on the state volume"})
+	} else {
+		add(passCheck("host.disk", fmt.Sprintf("%d bytes available", available)))
+	}
+
+	add(anyToolCheck(deps, "host.package-manager", []string{"apt-get", "dnf", "yum", "brew", "apk"}, "install a supported package manager or manage dependencies manually"))
+	add(anyToolCheck(deps, "host.service-manager", []string{"systemctl", "launchctl"}, "manage services manually on this platform"))
```
harshitsinghbhandari: PATH presence is not truthful service-manager support: `systemctl` commonly exists in containers/WSL where systemd is not PID 1, and this list can accept the wrong manager for the current OS. Select the manager by platform and perform a bounded, non-mutating usability probe; otherwise report unsupported/unavailable rather than pass.

backend/internal/haocli/doctor.go
```diff
@@ -0,0 +1,265 @@
...
+	add(diagnosticFromObservation(toolObservation(ctx, deps, "tool.gh", "gh", github, "--version")))
+	if github {
+		add(exitOnlyProbe(ctx, deps, "tool.gh.auth", "gh", []string{"auth", "status"}, "run gh auth login directly as the target user"))
+	} else {
+		add(DiagnosticCheck{ID: "tool.gh.auth", Severity: "info", Status: "disabled", Evidence: "GitHub workflow profile is not selected"})
+	}
+	harness := configString(object, "harness", "id")
+	add(diagnosticFromObservation(toolObservation(ctx, deps, "harness."+harness, harnessBinary(harness), true, "--version")))
+	if harness == "claude-code" {
+		add(exitOnlyProbe(ctx, deps, "harness.claude-code.auth", "claude", []string{"auth", "status", "--json"}, "run claude login directly as the target user"))
```
harshitsinghbhandari: This independently redefines harness authentication readiness even though the daemon doctor report below already carries the shared AO/harness truth. It also gives `claude auth status --json` the generic 2s budget, while `internal/doctor` deliberately uses 10s because cold Node startup routinely exceeds 2s. Consume the daemon check instead, or reuse the shared implementation/budget; otherwise healthy machines will oscillate between this `unknown` and the daemon's result. The GitHub auth probe above has the same dual-truth problem.

harshitsinghbhandari (commented):
## PR Review: #140

### Summary
The read-only shape and fixed loopback URL are directionally sound, and focused race/cross-build checks passed, but adversarial verification found multiple truthfulness and confidentiality failures. **Real verdict: CHANGES REQUESTED.**

### Security
The report-level redactor does not cover bearer credentials or secret values embedded as JSON, so daemon evidence/remediation can leak recursively in both human and JSON output.

### Consumer Impact
A timed-out daemon health request is rendered as `unhealthy` rather than `unknown`; an intentionally disabled absent daemon is rendered `unavailable` in status; and unknown daemon doctor levels default to `pass`. These are materially false distinctions for operators and automation. The new local auth probes also duplicate the daemon doctor truth with a shorter, divergent timeout.

### Testing
The focused HAO race suite passed. Adversarial tests reproduced bearer/embedded-JSON leakage and stable-ID collisions. Linux amd64/arm64 builds produced static x86-64/aarch64 ELF files with mode 0755. Existing daemon-wire tests use copied fake JSON rather than the real controller response, leaving that join unverified.

### Boundary Verification
**FAIL — 7 unverified boundaries:** exact `AO_RUN_FILE` producer → reconstructed discovery path; real daemon doctor DTO → HAO decoder; diagnostic strings → recursive human/JSON redaction; subprocess pipes → bounded capture; daemon check names → stable IDs; platform/service environment → service-manager result; selected auth requirements → duplicated local and daemon truth. No deferrals or overrides were present.

### CI and discussion
The head changed during review, so all findings and focused tests were rerun at `90314b9`. The latest CI run was still in progress at submission time; the preceding head had passed artifact and native platform jobs but failed lint. There was no existing PR discussion or inline review to reconcile.

### Recommendation
**Request Changes.** The boundary gate fails, and the findings below are actionable correctness/security issues rather than coverage-only suggestions.

BOUNDARY GATE FAILED: 7 unverified boundaries — see boundary-report.md (review artifact retained outside the read-only workspace).

backend/internal/haocli/doctor.go
harshitsinghbhandari: This creates a second Claude-auth definition alongside the daemon doctor consumed below. The shared doctor deliberately budgets 10s for the Node CLI and treats an unauthenticated harness as WARN/degraded; this uses the generic 2s budget and `exitOnlyProbe` converts nonzero into FAIL, so `hao doctor` exits 1 for a state canonical `ao doctor` does not call broken. Reuse the daemon result (or the shared probe implementation and semantics) rather than running a divergent check.

harshitsinghbhandari (commented):
## Independent architecture/security re-review — exact head `90314b9c6`

### Verdict
**CHANGES REQUESTED.** This review is deliberately pinned to `90314b9c655e15a78efc62f99534355c22e3ebba` versus `develop` `06bebe3a901c018a7819de7a29060a9483aa5c5e`; later amendments were excluded.

### Security and truthfulness
The fixed `127.0.0.1` construction, redirect refusal, response-size cap, and PID/service response matching confine daemon HTTP discovery to loopback without a DNS/redirect SSRF path. However, recursive output confidentiality fails: adversarial daemon evidence containing a bearer token, nested JSON token, pairing URI/passcode, and PEM-like private-key material survives into the report. Health transport timeouts are also reported as proven `unhealthy`, an absent intentionally-disabled daemon is `unavailable` in status but `disabled` in doctor, unknown daemon levels default to `pass`, and lossy normalization can publish duplicate check IDs.

### Auth probes and daemon-doctor reuse
The new exit-only `gh auth status` / `claude auth status --json` calls are profile/harness conditional and cancellation is bounded, but they independently redefine readiness already supplied by the shared daemon doctor. Claude auth gets the generic 2-second budget despite the shared implementation's deliberate 10-second cold-start budget, and a normal unauthenticated Claude state becomes `fail`, making `hao doctor` exit 1 where the canonical doctor defines unauthenticated harness state as WARN/degraded. A missing binary is simultaneously `unavailable` in the version check and `fail` in the auth check.

### Bounds, mutation, and packaging
Injected and real HTTP responses are bounded at 1 MiB, including the aggregation seam. Subprocess capture remains unbounded until process exit because truncation happens after `bytes.Buffer` allocation. Focused race tests pass. A before/after filesystem digest around real `status --strict` and `doctor --json` runs showed zero mutation. Linux amd64/arm64 builds are static x86-64/aarch64 ELF files with mode 0755; the exact-head artifact jobs passed. The exact head's lint job failed independently.

### Strict exits and partial failure
`status --strict` correctly leaves unknown observations non-fatal and fails proven unavailable desired state. Doctor endpoint failure remains partial (`ao.daemon` can stay healthy while `ao.doctor` is unknown). Those good properties do not compensate for the false auth FAIL and timeout/unavailable labels above.

### Boundary verification
**FAIL — 8 unverified boundaries, 0 deferrals, 0 overrides:** exact run-file override → discovery; real daemon doctor DTO → HAO decoder; daemon/tool strings → recursive human/JSON redaction; subprocess pipes → bounded capture; daemon names → stable IDs; selected auth state → shared daemon truth; platform environment → service-manager semantics; transport timeout → truthful observation state.

BOUNDARY GATE FAILED: 8 unverified boundaries — see boundary-report.md (ephemeral review artifact; workspace remained read-only).

backend/internal/haocli/doctor.go
```diff
@@ -0,0 +1,265 @@
...
+		if path, err := deps.Observer.LookPath(name); err == nil {
+			return passCheck(id, filepath.Base(path)+" is available")
+		}
+	}
+	return DiagnosticCheck{ID: id, Severity: "warning", Status: "unsupported", Evidence: "no supported implementation found", Remediation: remediation}
+}
+func exitOnlyProbe(ctx context.Context, deps Deps, id, binary string, args []string, remediation string) DiagnosticCheck {
+	path, err := deps.Observer.LookPath(binary)
+	if err != nil {
+		return DiagnosticCheck{ID: id, Severity: "error", Status: "fail", Evidence: binary + " is unavailable", Remediation: remediation}
```
harshitsinghbhandari: A missing executable is an inability to run the auth observation, not proof of failed authentication. The adjacent version observation reports the same fact as `unavailable`, while this reports `fail`, producing contradictory machine truth and incrementing doctor failures. Preserve `unavailable`/`unknown` here and reserve FAIL for a completed, authoritative negative auth result.

backend/internal/haocli/observe.go
harshitsinghbhandari: The command-output bound is applied only after `cmd.Run()` returns, so stdout/stderr can grow this buffer without limit during both version and the new auth probes. Use a writer that retains at most the configured prefix while continuing to drain/discard excess, and test a real child producing output beyond the cap.

backend/internal/haocli/observe.go
```diff
@@ -0,0 +1,247 @@
...
+	var health probePayload
+	if err := read("/healthz", &health); err != nil {
+		result.Evidence = "health probe failed: " + safeDiagnostic(err)
```
harshitsinghbhandari, line 194: A deadline/cancellation here returns the preinitialized `unhealthy` state. Failure to observe is `unknown`; it does not prove the daemon is unhealthy. Branch on `probeCtx.Err()` / `context.DeadlineExceeded`, matching `toolObservation`, and assert the `ao.daemon` status directly in tests.

backend/internal/haocli/doctor.go
harshitsinghbhandari: This boundary still accepts daemon-controlled strings and names without safe exhaustive projection: unknown `level` values default to pass, message/remediation bearer/JSON/pairing/PEM secrets survive report redaction, and lossy `stableID` normalization can collide. Exhaustively map levels, strengthen recursive string redaction, and preserve or validate collision-free daemon IDs with a real controller→decoder round trip.

backend/internal/haocli/observe.go
```diff
@@ -0,0 +1,249 @@
...
+func (systemObserver) Stat(path string) (FileObservation, error)      { return statFile(path) }
+func (systemObserver) Disk(path string) (uint64, error)               { return diskAvailable(path) }
+func (systemObserver) LookPath(name string) (string, error)           { return exec.LookPath(name) }
+func (systemObserver) ReadRunFile(path string) (*runfile.Info, error) { return runfile.Read(path) }
+func (systemObserver) ProcessAlive(pid int) bool                      { return processalive.Alive(pid) }
+
+func (systemObserver) Run(ctx context.Context, name string, args ...string) (string, error) {
+	cmd := exec.CommandContext(ctx, name, args...)
+	cmd.Stdin = nil
+	var out bytes.Buffer
```
harshitsinghbhandari: The real noisy-child reproduction emits 1 MiB on each pipe and completes, but this buffer retains the full 2 MiB before line 64 truncates the returned string. Replace it with a cap-retaining writer that returns the original byte count while draining/discarding excess; assert a real child well beyond the cap completes and only the prefix is retained.

backend/internal/haocli/status.go
```diff
@@ -0,0 +1,180 @@
...
+	path, object, err := loadConfig(deps, explicit)
+	if err != nil {
+		return StatusReport{}, err
+	}
+	stateRoot, err := deps.StateDir()
+	if err != nil {
+		return StatusReport{}, operationalError("resolve state root", err)
+	}
+	report := StatusReport{SchemaVersion: observationSchemaVersion, Machine: configString(object, "machine", "name"), Mode: configString(object, "mode"), ConfigPath: path, ConfigVersion: 1, HAOVersion: Version, DesiredAOVersion: configString(object, "components", "aoVersion")}
+	d := observeDaemon(ctx, deps.Observer, discoveryPath(stateRoot), deps.Timeout)
```
harshitsinghbhandari: Adversarial reproduction with `AO_RUN_FILE=/tmp/.../custom-daemon.json` records a read of `/tmp/.../running.json`. Resolve the exact path through `config.ResolveRunFilePath` (for doctor too) and assert the real env override reaches `ReadRunFile` unchanged.

backend/internal/haocli/status.go
```diff
@@ -0,0 +1,180 @@
...
+	desiredService := configBool(object, "service", "enabled")
+	report.Observations = append(report.Observations, Observation{ID: "ao.daemon", Status: d.State, Desired: desiredService, Evidence: d.Evidence})
```
harshitsinghbhandari, line 82: With local config (`service.enabled:false`) and no discovery file, `status --json --strict` emits `ao.daemon: unavailable, desired:false`, while doctor emits disabled. Normalize the status observation to `disabled` so the commands expose one truthful state.

backend/internal/haocli/observe.go
```diff
@@ -0,0 +1,249 @@
...
+			return errors.New("response exceeds size limit")
+		}
+		if err := json.Unmarshal(body, target); err != nil {
+			return fmt.Errorf("malformed response")
+		}
+		return nil
+	}
+	var health probePayload
+	if err := read("/healthz", &health); err != nil {
+		result.Evidence = "health probe failed: " + safeDiagnostic(err)
```
harshitsinghbhandari, line 194: A `context.DeadlineExceeded` health request leaves the preinitialized state as `unhealthy`. A timeout proves only that observation failed; map deadline/cancellation to `unknown`, consistent with tool probes, and test the daemon observation directly.

backend/internal/haocli/doctor.go
```diff
@@ -0,0 +1,267 @@
...
+	} else {
+		for _, check := range d.Doctor.Checks {
+			status, severity := "pass", "info"
+			if check.Level == "WARN" {
+				status, severity = "warn", "warning"
+			}
+			if check.Level == "FAIL" {
+				status, severity = "fail", "error"
+			}
+			add(DiagnosticCheck{ID: "ao.doctor." + stableID(check.Name), Severity: severity, Status: status, Evidence: check.Message, Remediation: check.Remediation})
```
harshitsinghbhandari: A real daemon-shaped response with level `MAYBE`, names `a b`/`a-b`, bearer + nested JSON + pairing + PEM material reproduces three failures here: unknown level becomes pass, IDs collide, and all secrets survive. Exhaustively project PASS/WARN/FAIL, use collision-free IDs, recursively redact strings, and round-trip the real controller DTO into this decoder.

backend/internal/haocli/doctor.go
```diff
@@ -0,0 +1,267 @@
...
+	if available, diskErr := deps.Observer.Disk(stateRoot); diskErr != nil {
+		add(errorCheck("host.disk", "disk availability probe failed"))
+	} else if available < 512*1024*1024 {
+		add(DiagnosticCheck{ID: "host.disk", Severity: "error", Status: "fail", Evidence: fmt.Sprintf("%d bytes available", available), Remediation: "free at least 512 MiB on the state volume"})
+	} else {
+		add(passCheck("host.disk", fmt.Sprintf("%d bytes available", available)))
+	}
+
+	add(anyToolCheck(deps, "host.package-manager", []string{"apt-get", "dnf", "yum", "brew", "apk"}, "install a supported package manager or manage dependencies manually"))
+	add(anyToolCheck(deps, "host.service-manager", []string{"systemctl", "launchctl"}, "manage services manually on this platform"))
```
harshitsinghbhandari: On a simulated Darwin host with only `/usr/bin/systemctl`, this reports pass. Choose the manager by platform and run a bounded non-mutating usability probe (including systemd-not-PID-1); PATH presence alone is not support.

backend/internal/haocli/doctor.go
```diff
@@ -0,0 +1,267 @@
...
+	add(diagnosticFromObservation(toolObservation(ctx, deps, "tool.gh", "gh", github, "--version")))
+	if github {
+		add(exitOnlyProbe(ctx, deps, "tool.gh.auth", "gh", []string{"auth", "status"}, "run gh auth login directly as the target user"))
+	} else {
+		add(DiagnosticCheck{ID: "tool.gh.auth", Severity: "info", Status: "disabled", Evidence: "GitHub workflow profile is not selected"})
+	}
+	harness := configString(object, "harness", "id")
+	add(diagnosticFromObservation(toolObservation(ctx, deps, "harness."+harness, harnessBinary(harness), true, "--version")))
+	if harness == "claude-code" {
+		add(exitOnlyProbe(ctx, deps, "harness.claude-code.auth", "claude", []string{"auth", "status", "--json"}, "run claude login directly as the target user"))
```
harshitsinghbhandari: This duplicates the daemon's canonical `claude-auth` check: the reproduction publishes both IDs with potentially contradictory outcomes. It also uses the generic 2s timeout versus the shared doctor's deliberate 10s Node cold-start budget and turns ordinary unauthenticated WARN into FAIL/exit 1. Consume the daemon-owned truth instead of rerunning auth locally.

backend/internal/haocli/doctor.go
```diff
@@ -0,0 +1,267 @@
...
+		if path, err := deps.Observer.LookPath(name); err == nil {
+			return passCheck(id, filepath.Base(path)+" is available")
+		}
+	}
+	return DiagnosticCheck{ID: id, Severity: "warning", Status: "unsupported", Evidence: "no supported implementation found", Remediation: remediation}
+}
+func exitOnlyProbe(ctx context.Context, deps Deps, id, binary string, args []string, remediation string) DiagnosticCheck {
+	path, err := deps.Observer.LookPath(binary)
+	if err != nil {
+		return DiagnosticCheck{ID: id, Severity: "error", Status: "fail", Evidence: binary + " is unavailable", Remediation: remediation}
```
harshitsinghbhandari: When the executable is absent, the version check says `unavailable` but this auth check says `fail`, creating contradictory truth and incrementing failures. Preserve unavailable/unknown for an unexecutable probe; reserve fail for an authoritative completed negative result.

harshitsinghbhandari (commented):
## Exact-head architecture/security review — `0e50cec9`

### Verdict
**CHANGES REQUESTED.** This review is pinned to `0e50cec957002756645faf591e73218c77cbb4d6` versus `develop` `06bebe3a901c018a7819de7a29060a9483aa5c5e`; later PR commits are intentionally excluded. All eight prior findings were independently rerun.

### Eight-finding verification
1. **FAIL — bounded subprocess draining:** a real 2 MiB noisy child completes, but source retains all output in `bytes.Buffer` and truncates only after exit.
2. **FAIL — exact discovery path:** with `AO_RUN_FILE=.../custom-daemon.json`, status reads sibling `running.json`.
3. **FAIL — disabled truth:** an absent intentionally-disabled daemon is `unavailable` in status, while doctor says `disabled`.
4. **FAIL — timeout truth:** a health deadline is published as `unhealthy`, not `unknown`.
5. **FAIL — real doctor projection:** unknown daemon levels default to `pass`; the production test copies JSON rather than using the real controller DTO.
6. **FAIL — redaction/IDs:** bearer, pairing/passcode, PEM, and embedded-JSON secrets leak; `a b` and `a-b` publish duplicate IDs.
7. **FAIL — service manager:** Darwin with only `systemctl` on PATH is reported supported/pass without a usable platform-correct manager.
8. **FAIL — single readiness truth:** HAO emits local `harness.claude-code.auth` alongside daemon `claude-auth`, with a divergent 2s budget and FAIL semantics.

### Verified properties
Daemon URLs are constructed only from fixed `http://127.0.0.1:<validated-port>` paths; redirects are refused, bodies are capped at 1 MiB at both transport and aggregation layers, and health responses must match AO service/PID. No DNS-controlled host or redirect SSRF path was found. `status --strict` keeps unknown non-fatal and fails proven unavailable desired state. Focused race tests pass. Before/after filesystem digests for real status/doctor runs prove zero mutation. Linux amd64/arm64 builds are static x86-64/aarch64 ELF executables with mode 0755. Exact-head CI passed build-test, artifacts, native platforms, API drift, container, and secret scan; lint failed.

### Boundary gate
**FAIL — 7 unverified seams, 0 deferrals, 0 overrides:** subprocess output→capture; env override→run-file reader; config+discovery→truthful state/strict exit; daemon DTO→HAO projection; diagnostic strings/names→safe stable report; platform→usable service manager; selected auth→single daemon-owned readiness truth.

BOUNDARY GATE FAILED: 7 unverified boundaries — see boundary-report.md (ephemeral review artifact; workspace remained read-only).

harshitsinghbhandari (commented):
## Exact-head architecture/security review — `cc1d584ae94d635f5c5985bd964e0de3ffddd95d`

### Verdict
**CHANGES REQUESTED.** This review is pinned exclusively to `cc1d584ae94d635f5c5985bd964e0de3ffddd95d` versus `develop` `06bebe3a901c018a7819de7a29060a9483aa5c5e`.

### Must-fix findings

1. `backend/internal/haocli/doctor.go:196-197` — **PATH presence is still reported as a usable service manager.** The new OS selection correctly requires `launchctl` on Darwin and rejects a stray `systemctl`, but Linux still passes whenever `systemctl` is installed even when systemd is not the active manager (containers, WSL, chroots). Run a bounded, non-mutating usability probe and report unsupported/unknown when it cannot communicate with the manager; the current test proves name selection only.

2. `backend/internal/haocli/doctor.go:137-147` — **Unknown daemon doctor levels do not fail closed at the process boundary.** An unrecognized level becomes `status:"unknown"`, but `add` counts only `fail` and `error`, so the real DTO compatibility test explicitly accepts exit 0. The daemon wire is unversioned; a new or corrupt level must increment `executionFailures` (or otherwise make doctor exit 1), not produce a successful diagnostic run.

3. `backend/internal/haocli/doctor.go:147` — **The collision fix makes IDs unique but not stable.** Prefixing the daemon check with its list position means inserting or reordering one daemon check renames every following ID. Machine consumers cannot safely key on these advertised stable IDs. Derive identity from the daemon-defined stable name and add deterministic collision disambiguation that does not depend on report order.

### Boundary verification
**FAIL — 3 unverified boundaries, 0 deferrals, 0 overrides:** platform → actually usable service manager; unversioned daemon level → fail-closed exit; daemon stable check name → collision-free stable HAO ID.

The remaining requested seams passed at this exact head: bounded subprocess draining; exact custom `AO_RUN_FILE`; disabled/unavailable and timeout/unknown distinctions; real controller DTO decoding; recursive bearer/pairing/PEM/embedded-JSON redaction; daemon-owned readiness truth; fixed-loopback SSRF and redirect confinement; strict failure for proven unhealthy desired state; read-only workspace behavior; focused race tests; Linux/Darwin/Windows cross-compiles; and checksum-valid executable Linux x64/arm64 AO/HAO artifacts. Darwin correctly rejects `systemctl` and requires `launchctl`. Current CI has every completed job green; `build-test` was still running at the final observation.

BOUNDARY GATE FAILED: 3 unverified boundaries — see the Boundary Verification section above.

_GitHub does not allow the PR author to submit a `REQUEST_CHANGES` event on their own PR, so this formal commit-attached review uses state COMMENTED; the verdict above is CHANGES REQUESTED._

harshitsinghbhandari (commented):
## Final exact-head architecture/security review — `03d0818670d21418366f2d4fb1d986325a2c365b`

### Verdict
**CHANGES REQUESTED.** This review evaluates only `03d0818670d21418366f2d4fb1d986325a2c365b` against `develop` `06bebe3a901c018a7819de7a29060a9483aa5c5e`.

### Must-fix finding

`backend/internal/haocli/doctor.go:138-154` — **The daemon check IDs remain collision-prone.** Distinct names are disambiguated with only the first 32 bits of SHA-256, without checking the final emitted ID. I reproduced two valid distinct names, `a<U+F260>b` and `a<U+1276D>b`, that both normalize to `a-b` and both produce suffix `5b7be23d`, yielding the same `ao.doctor.a-b-5b7be23d` ID. The duplicate-name path also repeats the same `ao.doctor.invalid-duplicate-<hash>` ID for the third and later occurrence of one name. This violates the collision-free machine contract even though ordinary insertion/reordering is now stable. Detect collisions on the final emitted IDs and fail the daemon projection once with a single stable aggregate error, or use a lossless encoding of the original daemon name; add adversarial tests for hash collision and three identical names.

### Reverification
The other two blockers are fixed: `systemctl show-environment` / `launchctl print system` are bounded, non-mutating, platform-correct usability probes with timeout and unusable-manager truth; unknown daemon levels now become execution failures and make `hao doctor` exit 1. Ordinary normalized-name collisions are order-stable under insertion/reordering.

All previously passing boundaries remain verified: bounded subprocess draining; exact custom `AO_RUN_FILE`; disabled/unavailable and timeout/unknown truth; real DTO decoding; recursive bearer/pairing/PEM/embedded-JSON redaction; daemon-owned readiness; loopback-only SSRF/redirect confinement; strict exits; zero workspace mutation; focused race tests; Linux/Darwin/Windows cross-compiles; statfs builds on supported platforms; and checksum-valid executable Linux x64/arm64 AO/HAO artifacts. Every completed CI job was green at final observation; Windows native and build-test were still running.

### Boundary verification
**FAIL — 1 unverified boundary, 0 deferrals, 0 overrides:** daemon check name → collision-free, order-stable machine ID.

BOUNDARY GATE FAILED: 1 unverified boundary — see the Boundary Verification section above.

_GitHub does not permit the PR author to submit a `REQUEST_CHANGES` event on their own PR, so this commit-attached formal review uses state COMMENTED; the verdict is CHANGES REQUESTED._

harshitsinghbhandari (commented):
## Final exact-head architecture/security review — `28193684d3b0af4740e9fd45f636dee3d53d621f`

### Verdict
**CLEAN.** This review evaluates only `28193684d3b0af4740e9fd45f636dee3d53d621f` against `develop` `06bebe3a901c018a7819de7a29060a9483aa5c5e`.

The final stable-ID boundary is fixed. The reproduced Unicode pair `a<U+F260>b` and `a<U+1276D>b` now receives distinct lossless base64url suffixes. IDs remain stable under insertion and reordering, including normalized-name collisions and empty normalized names. Two, three, or more identical raw names are omitted and replaced by exactly one deterministic `ao.projection.duplicate-doctor-names` execution failure, and `hao doctor` exits 1.

### Boundary verification
**PASS — 0 unverified boundaries, 0 deferrals, 0 overrides.**

Regression verification also passed for bounded subprocess draining; exact custom `AO_RUN_FILE`; disabled/unavailable and timeout/unknown truth; unknown daemon levels failing closed through the real DTO; recursive bearer/pairing/PEM/embedded-JSON redaction; daemon-owned readiness; bounded platform-correct `systemctl`/`launchctl` usability; fixed-loopback SSRF and redirect confinement; strict exits; zero workspace mutation; race; supported-platform statfs compilation; and Linux/Darwin/Windows cross-builds. Linux x64/arm64 AO and HAO artifacts are checksum-valid, executable, static binaries of the expected architectures. All completed CI jobs were green; build-test remained in progress at the immediate pre-submit observation.

No actionable architecture, security, compatibility, or boundary findings remain.

_GitHub does not permit the PR author to approve their own PR, so this commit-attached formal review uses state COMMENTED; the verdict is CLEAN._

PR #141, opened 2026-08-29
fix: block browser routes at VM gateway

0xkaushik-ai:
## Summary
- block /api/v1/browser at the public VM gateway
- update route policy contract and regression coverage

## Tests
- go test ./internal/vmgateway
- full backend suite: unrelated existing failure in internal/adapters/agent/fake/TestFullLifecycleSpawnToTermination (missing events.log)


backend/internal/vmgateway/proxy_test.go
```diff
@@ -205,6 +205,7 @@ func TestGateway_BlockedRoutes_NeverReachDaemon(t *testing.T) {
 		"/api/v1/mobile/status",
 		"/api/v1/mobile/devices",
 		"/api/v1/mobile",
+		"/api/v1/browser/status",
```
harshitsinghbhandari, line 208: This only exercises the blocked browser path with a valid token, so it does not lock down the security-critical middleware ordering this route relies on. Please add a real-handler case with no credential (and ideally an invalid credential) asserting 404—not 401—and zero daemon calls. That proves `denyByDefault` runs before both hosted-token and pair-passcode authentication, rather than merely proving the helper rejects the path.

harshitsinghbhandari (changes requested):
## PR Review: #141

### Summary
The segment-boundary implementation is correct: `/api/v1/browser` and descendants are blocked while `/api/v1/browserapp` remains proxyable, and the shared handler preserves REST/SSE/`/mux` behavior. The route-policy contract matches the runtime prefix list.

### Security
The new end-to-end regression only exercises a valid-token request. It therefore proves 404 plus zero daemon calls, but not the load-bearing 404-before-auth ordering for a missing/invalid credential. That ordering is part of the gateway's route-concealment boundary and is shared by hosted-token and pair-passcode modes.

### Boundary Verification
**FAIL — 1 unverified boundary.**

- `denyByDefault(isProxyablePath("/api/v1/browser/status"))` → hosted/pair authentication middleware: the contract is “blocked route returns 404 before credential processing and makes zero upstream daemon calls.” The existing test covers the valid-token/zero-upstream half, but no real-handler test covers the unauthenticated ordering for the newly blocked family.

No contract-shape drift was found. The HAO JSON fixture and runtime list remain synchronized.

### CI
GitHub reports no checks for head `c01def24`. I independently ran `go test ./internal/vmgateway` against that commit; it passed.

### Recommendation
**Request Changes**

Add a real-handler regression for an unauthenticated (and ideally invalid-token) `/api/v1/browser/status` request that asserts 404 and zero daemon calls.

harshitsinghbhandari (approved):
## PR Review: #141

### Summary
Clean at exact head d66b542dc6330cd125c2e2fd5004925aec124284. The new missing-token and invalid-token real-handler cases both prove /api/v1/browser/status returns 404 before hosted authentication and makes zero daemon calls; the existing valid-token regression proves the same route remains blocked after successful authentication.

### Boundary Verification
PASS — 0 unverified boundaries. The hosted and pair gateways share newHandler, placing denyByDefault outside the credential middleware in both modes. Segment-boundary coverage blocks /api/v1/browser and descendants while allowing /api/v1/browserapp; existing REST, SSE, and /mux routes remain covered. The LAN listener independently blocks the browser family before LAN auth, and the HAO contract fixture matches the gateway runtime prefix list and boundary behavior.

### Verification
- go test ./internal/vmgateway -count=1
- Focused internal/httpd LAN tests
- GitHub currently reports no checks for this head.

### Recommendation
APPROVE. The prior 404-before-auth coverage gap is closed with real assembled-handler tests, and no new correctness, security, compatibility, or contract issues were found.

PR #142, opened 2026-08-29
fix: refresh pairing probe sessions after misses

0xkaushik-ai:
## Summary
- capture pair-mode leaf certificates over a short-lived raw TLS socket, outside Chromium
- remove generated Electron probe partitions and their process-lifetime growth
- canonicalize DNS, IPv4, and bracketed/bare IPv6 host keys for pin lookup
- keep Electron certificate verification focused on steady-state pinned traffic
- add real TLS regression coverage for retries, concurrency, timeout/abort cleanup, and host normalization

## Tests
- `npm exec vitest run src/main/paired-machine-probe.test.ts src/main/paired-machines.test.ts src/main/paired-machine-cert.test.ts`
- `npm run typecheck`
- `npm run typecheck:e2e`

## Notes
The capture socket sends no HTTP request or credentials. It reads the presented leaf certificate, computes the existing SHA-256 fingerprint format, and closes. The user must still compare and explicitly pin that fingerprint before Electron traffic can be accepted.

### DIFF
--- a/frontend/src/main.ts
+++ b/frontend/src/main.ts
@@ -2414,8 +2414,11 @@ ipcMain.handle("pairedMachines:list", () => pairedMachines().list());
 ipcMain.handle("pairedMachines:refresh", () => pairedMachines().refresh());
 ipcMain.handle(
   "pairedMachines:probeFingerprint",
-  (_event, address: string, port: number) =>
-    pairedMachines().probeFingerprint(address, port),
+  (_event, address: string, port: number, requestId?: string) =>
+    pairedMachines().probeFingerprint(address, port, requestId),
+);
+ipcMain.on("pairedMachines:cancelFingerprintProbe", (_event, requestId: string) =>
+  pairedMachines().cancelFingerprintProbe(requestId),
 );
 ipcMain.handle("pairedMachines:getPinnedFingerprint", (_event, id: string) =>
   pairedMachines().getPinnedFingerprint(id),
--- a/frontend/src/main.ts
+++ b/frontend/src/main.ts
@@ -2331,25 +2331,23 @@ const LOCAL_MACHINE_NAME =
  * distinct box ever probed, which is what actually needs isolating from
  * `session.defaultSession`.
  *
- * The cost: a probe of an unpinned host is always denied, by design (same as
- * above), so reusing that host's session for a second probe can hit its own
- * cached rejection and fail to capture anything -- not only when the box's
- * certificate has actually changed between the two attempts. The realistic
- * trigger is a retry after an abandoned or failed pairing attempt (compare
- * the fingerprint, close the dialog without pinning, try again), which can
- * come back with a misleading "no certificate could be retrieved" until the
- * app restarts. Re-pairing an already-registered box is unaffected: a pinned
- * host verifies and accepts on `session.defaultSession`, which this probe
- * session never touches. Tracked as a follow-up: #121.
  */
+const pairProbeSessionGenerations = new Map<string, number>();
+
 function pairProbeNetFetch(
   input: RequestInfo | URL,
   init?: RequestInit,
 ): Promise<Response> {
   const hostname = new URL(String(input)).hostname;
-  const probeSession = session.fromPartition(`pair-probe-${hostname}`);
+  const generation = pairProbeSessionGenerations.get(hostname) ?? 0;
+  const probeSession = session.fromPartition(`pair-probe-${hostname}-${generation}`);
   probeSession.setCertificateVerifyProc(pairedMachines().verifyCertificate);
   return probeSession.fetch(String(input), init);
+}
+
+function resetPairProbeSession(address: string): void {
+  const current = pairProbeSessionGenerations.get(address) ?? 0;
+  pairProbeSessionGenerations.set(address, current + 1);
 }
 
 let pairedMachinesController: ReturnType<
@@ -2366,6 +2364,7 @@ function pairedMachines(): ReturnType<typeof createPairedMachinesController> {
     // fetch would bypass the session (and the pin) entirely.
     netFetch: (input, init) => net.fetch(String(input), init),
     probeNetFetch: pairProbeNetFetch,
+    onProbeMiss: resetPairProbeSession,
   });
   return pairedMachinesController;
 }
--- a/frontend/src/main.ts
+++ b/frontend/src/main.ts
@@ -2364,7 +2364,7 @@ function pairedMachines(): ReturnType<typeof createPairedMachinesController> {
     // fetch would bypass the session (and the pin) entirely.
     netFetch: (input, init) => net.fetch(String(input), init),
     probeNetFetch: pairProbeNetFetch,
-    onProbeMiss: resetPairProbeSession,
+    onProbeComplete: resetPairProbeSession,
   });
   return pairedMachinesController;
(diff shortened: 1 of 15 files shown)


0xkaushik-ai:
@harshitsinghbhandari

frontend/src/main/paired-machines.ts
```diff
@@ -584,6 +585,7 @@ export function createPairedMachinesController(deps: PairedMachinesControllerDep
 			}
 
 			const fingerprint = presented.get(address);
+			if (!fingerprint) deps.onProbeMiss?.(address);
```
harshitsinghbhandari: A successful capture must also retire this probe session. The verify proc intentionally rejects the unpinned certificate even when it records a fingerprint, so Chromium caches a rejection in this session. If the user closes the comparison dialog and retries, this condition skips rotation; the second probe reuses the poisoned session and returns “No certificate,” and only that miss rotates for a third attempt. Please model the actual capture → abandon → retry flow and test it with the session-cache stand-in rather than only spying on this callback.

frontend/src/main.ts
```diff
@@ -2331,27 +2331,25 @@ const LOCAL_MACHINE_NAME =
...
+const pairProbeSessionGenerations = new Map<string, number>();
+
 function pairProbeNetFetch(
   input: RequestInfo | URL,
   init?: RequestInit,
 ): Promise<Response> {
   const hostname = new URL(String(input)).hostname;
-  const probeSession = session.fromPartition(`pair-probe-${hostname}`);
+  const generation = pairProbeSessionGenerations.get(hostname) ?? 0;
+  const probeSession = session.fromPartition(`pair-probe-${hostname}-${generation}`);
```
harshitsinghbhandari: This restores the process-lifetime growth the preceding comment says must be avoided: every miss advances the generation, while `session.fromPartition` retains every generated partition because there is no destroy API. Repeated retries, unreachable hosts, canceled dialog work, and losing race candidates can retain one more Chromium Session each time. Please use a lifecycle that bounds retained partitions and test that N failures/cancellations do not create O(N) sessions.

frontend/src/main.ts
```diff
@@ -2331,27 +2331,25 @@ const LOCAL_MACHINE_NAME =
...
   probeSession.setCertificateVerifyProc(pairedMachines().verifyCertificate);
   return probeSession.fetch(String(input), init);
 }
 
+function resetPairProbeSession(address: string): void {
+  const current = pairProbeSessionGenerations.get(address) ?? 0;
```
harshitsinghbhandari: Normalize this key exactly as `pairProbeNetFetch` does. For bare IPv6, this callback receives `2001:db8::1`, but `new URL(...).hostname` reads `[2001:db8::1]`; the generation is written under one key and read under another, so IPv6 retries stay on generation 0. Please share one canonical-host helper and cover IPv4, DNS normalization, and bare IPv6.

harshitsinghbhandari (changes requested):
## PR Review: #142

### Summary
The callback is wired through, but the actual capture → abandon → retry flow remains broken, IPv6 generations use inconsistent keys, and failed/canceled probes cause unbounded Electron-session growth.

### Boundary Verification
**FAIL — 3 blocking boundaries:** successful certificate capture does not rotate the intentionally poisoned session; raw controller addresses do not round-trip to URL-normalized session keys; and mocked unit tests never exercise the real Electron partition/cache seam. No deferrals or overrides apply.

### Testing / CI
The tests assert callback calls rather than miss → retry and success → abandon → retry behavior. They omit IPv6, concurrency/cancellation, stale completions, resource bounds, and packaged-Electron verification. GitHub reports gitleaks, Frontend, and Go as `action_required`, so no CI checks have executed.

### Recommendation
**CHANGES REQUESTED**

The second attempt can still fail after a successful capture, IPv6 never rotates, and retries can retain O(N) Chromium sessions for process lifetime.

0xkaushik-ai:
@harshitsinghbhandari I replaced the partition-generation approach in 0e74191d.

The capture path now uses a short-lived raw TLS socket and never enters Chromium, so retries cannot poison the default session and failed/cancelled/repeated probes retain no Electron Sessions. The socket is destroyed on capture, error, timeout, and abort.

The same canonical-host helper now normalizes DNS, IPv4, and bracketed/bare IPv6 before pin lookup. Capture returns its fingerprint directly, removing the shared pending/presented maps and stale-completion race.

Coverage now exercises a real local TLS server for repeated captures, concurrent captures, timeout cleanup, and abort cleanup, plus controller retry/concurrency behavior and bracketed Electron IPv6 host lookup. Focused tests, project typecheck, and E2E typecheck pass. CI is running on the pushed commit.

harshitsinghbhandari (dismissed):
## Strict review of 0e74191d7e0075f95e5eb9ac6f119556f6914fc0

This supersedes my review of `3b7576f`. I independently reproduced the old scenarios. The raw TLS capture fixes capture, abandon, retry after a successful fingerprint; generated Electron probe partitions are now zero; and DNS, IPv4, plus bracketed and bare IPv6 keys normalize through one helper.

### High: abandoned pairing races do not cancel raw TLS probes, so resource use is not bounded by the live UI operation

`capturePairFingerprint` supports an `AbortSignal`, but `paired-machines.ts:550-556` never accepts or passes one. The real bridge preserves that gap: `preload.ts:450-451` and `main.ts:2415-2418` expose only `(address, port)`. Meanwhile `racePairAddresses` treats renderer abort as result cancellation only. It cannot stop probes it already started. Closing the dialog or starting another attempt therefore leaves every raw TLS socket alive until the fixed 8-second timeout. The pairing-string parser also places no cap on address count, so one attempt can start an arbitrary number of sockets, and repeated cancel/retry multiplies that live set.

Please propagate cancellation across the renderer, IPC, controller, and `capturePairFingerprint`, and cap candidate fanout or otherwise enforce a hard aggregate bound. Add a real seam test that starts non-handshaking peers, cancels the user race, and proves all server-side connections close promptly rather than only after timeout. The current abort test calls `capturePairFingerprint` directly, so it does not exercise the broken production path.

### Boundary verification

**FAIL: 1 unverified boundary out of 8 reviewed.** The failing seam is `AddPairedMachineDialog.tsx:160-192` to `pair-race.ts:139-160` to `preload.ts:450-451` to `main.ts:2415-2418` to `paired-machines.ts:550-556` to `paired-machine-probe.ts:91-107`. Contract: abandoning a user attempt must cancel and release its in-flight transports. Actual: cancellation stops UI completion but does not reach the transport. No deferrals apply.

The other seven reviewed seams pass by source trace and tests: address canonicalization into TLS connect; DER leaf into SHA-256 fingerprint; capture result into controller result; stateless repeated capture; concurrent capture isolation; stored raw host into canonical pin lookup and Electron verifier; and persisted record reload. The packaged Electron network-service cache boundary was not exercised locally, but capture no longer enters Chromium at all, and no probe `Session` or partition remains in the code.

### Tests and integration

At the exact PR head: 41 focused tests passed; `typecheck` and `typecheck:e2e` passed. On a clean synthetic merge into current `develop` (`504324c99c56bdc7244239dc3db500962c37ea08`): 70 pairing and dialog tests passed, plus both typechecks. GitHub currently reports all 7 checks passing and the merge is conflict-free.

**Verdict: CHANGES REQUESTED.**

PR #143, opened 2026-08-29
test: keep renderer API client mocks complete

### LINKED ISSUE #104: api-client wholesale mocks lack base-url exports (17 files)
17 renderer test files vi.mock lib/api-client wholesale without getApiBaseUrl/subscribeApiBaseUrl. Any of them that ever mounts CreateProjectFlow (or anything else using useSyncExternalStore over the api base url) fails with an opaque error. PR #98 fixed the three that mount it today (Sidebar, board-empty-states, CommandPalette). Consider a shared mock helper.

0xkaushik-ai:
## Summary
- add `getApiBaseUrl` and `subscribeApiBaseUrl` to renderer API-client mocks
- keep wholesale mocks aligned with the API client contract
- cover all current renderer component, hook, integration, and orchestration mocks

Closes #104

## Validation
- `npm exec vitest run src/renderer/components src/renderer/hooks src/renderer/__tests__/integration` (108 files, 1630 tests)
- `npm run typecheck`
- `git diff --check

### DIFF
--- a/frontend/src/renderer/__tests__/integration/pr-hydration.test.tsx
+++ b/frontend/src/renderer/__tests__/integration/pr-hydration.test.tsx
@@ -12,6 +12,8 @@ vi.mock("../../lib/api-client", () => ({
 	apiClient: { GET: getMock, POST: vi.fn() },
 	apiErrorMessage: (e: unknown) => (e instanceof Error ? e.message : "error"),
 	hasTrustedApiBaseUrl: () => true,
+	getApiBaseUrl: () => "http://127.0.0.1:3001",
+	subscribeApiBaseUrl: () => (() => {}),
 }));
 
 vi.mock("@tanstack/react-router", async (importOriginal) => {
--- a/frontend/src/renderer/__tests__/integration/pr-hydration.test.tsx
+++ b/frontend/src/renderer/__tests__/integration/pr-hydration.test.tsx
@@ -8,12 +8,11 @@ import type { ReactNode } from "react";
 // on the session list flow through the shared workspace cache into the board.
 const { getMock, navigateMock } = vi.hoisted(() => ({ getMock: vi.fn(), navigateMock: vi.fn() }));
 
-vi.mock("../../lib/api-client", () => ({
+vi.mock("../../lib/api-client", async (importOriginal) => ({
+	...(await importOriginal<typeof import("../../lib/api-client")>()),
 	apiClient: { GET: getMock, POST: vi.fn() },
 	apiErrorMessage: (e: unknown) => (e instanceof Error ? e.message : "error"),
 	hasTrustedApiBaseUrl: () => true,
-	getApiBaseUrl: () => "http://127.0.0.1:3001",
-	subscribeApiBaseUrl: () => (() => {}),
 }));
 
 vi.mock("@tanstack/react-router", async (importOriginal) => {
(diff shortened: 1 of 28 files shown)


frontend/src/renderer/__tests__/integration/pr-hydration.test.tsx
```diff
@@ -12,6 +12,8 @@ vi.mock("../../lib/api-client", () => ({
 	apiClient: { GET: getMock, POST: vi.fn() },
 	apiErrorMessage: (e: unknown) => (e instanceof Error ? e.message : "error"),
 	hasTrustedApiBaseUrl: () => true,
+	getApiBaseUrl: () => "http://127.0.0.1:3001",
```
harshitsinghbhandari: Non-blocking: this identical stable-store pair is now repeated across 28 mocks. Please consider a shared renderer test helper (or a safe `importOriginal` partial mock where module initialization is acceptable) so future `api-client` exports cannot drift across whole-module mocks. The current constant snapshot/no-op subscription is correct for this suite.

harshitsinghbhandari (approved):
## PR Review: #143

### Summary
This test-only change restores the two base-URL store exports required by transitive renderer consumers in every affected whole-module mock. The fixed snapshot plus no-op subscription is a valid static `useSyncExternalStore` implementation for suites that do not exercise machine switching, and the diff has no production-code impact.

### Boundary Verification
PASS — the production `api-client` export contract → whole-module Vitest mocks → transitive React/React Query consumers is satisfied in the affected suites. The dedicated production-store tests remain responsible for URL-change notifications; these component mocks intentionally represent a stable local daemon. No deferrals or overrides.

### Testing and CI
Verified in an isolated clone at `158fc360`: 108 targeted test files / 1,630 tests passed, frontend typecheck passed, and `git diff --check` passed. GitHub currently reports no check runs or commit statuses for this external-fork head, so hosted CI evidence is unavailable.

### Optional refactoring
The repeated two-export stubs remain a maintenance hotspot. Consolidating the stable local-daemon store in shared test infrastructure (or using a safe partial-module mock where appropriate) would prevent the next production export addition from requiring another repository-wide sweep. This is non-blocking because the current contract and semantics are correct.

### Verdict
**APPROVE / CLEAN**

harshitsinghbhandari (approved):
## PR Review: #143

### Summary
The `importOriginal` partial-mock refactor is proportionate and safe at `a0e4935440e11f1550897cb68509296b7393856c`. It preserves the production export surface while retaining only the per-suite API method/error overrides, and the three explicit loopback base-URL overrides remain limited to suites whose folder-picker assertions specifically require the local-machine path.

### Mock and boundary verification
PASS — importing `api-client` initializes only its in-memory URL/status stores and the OpenAPI client. Network fetches, gateway-token access, telemetry, and Sentry reporting remain call-driven; the overridden `apiClient` prevents those paths from executing in these suites. The real base-URL store has a stable snapshot and proper subscribe/unsubscribe semantics for `useSyncExternalStore`, while tests that exercise EventSource rebinding keep their deliberate wholesale stateful mocks. The remaining narrow wholesale mocks are purpose-specific blob-URL or transport seams and are not omitted broad mocks. No shared mutable mock state or production side-effect leak was found.

### Verification
- `git diff --check 06bebe3a...a0e4935`: passed.
- Exact-SHA renderer run: 104/108 suites and 1,546 tests passed. Four suites could not resolve dependencies because the available reusable dependency tree predates `react-arborist` / `@react-symbols/icons`; a clean local install is blocked by the environment npm git-dependency policy. This is an environment limitation, not a test assertion failure. The changed suites affected by it were `FileTree.test.tsx` and `SessionView.test.tsx`; their refactors were inspected directly.
- Typecheck was not independently reproducible for the same incomplete-dependency reason.
- GitHub reports no check runs for this external-fork head.

### Verdict
**APPROVE / CLEAN**

Issue #136, opened 2026-08-28
Move agent sessions between machines


harshitsinghbhandari:
## User goal

Let a user hand off an active agent session directly from machine A to machine B through an HAO-assisted checkpoint/export and restore/resume flow.

The handoff should preserve portable session context and workspace state while safely excluding or re-establishing credentials and recreating machine-local runtime state on the destination.

## Acceptance outcomes

- A user can initiate a direct machine-to-machine handoff and resume the session on the destination machine.
- Portable agent/session context and relevant workspace state survive the handoff with integrity checks and actionable failure reporting.
- Secrets and credentials are not embedded in the portable checkpoint; destination credentials are re-established safely.
- Machine-local runtime state is identified as non-portable and recreated or reconnected on the destination.
- The source session is left in a clear, recoverable state if export, transfer, or restore fails.

decebal:
Harshit — Hosted AO #136 needs machine A → B handoff while preserving portable session/workspace state, excluding secrets, and leaving source recoverable when transfer fails. I build AllSource, an event-sourced agent-memory and provenance layer. I can help test an export manifest whose state transitions, integrity checks, exclusions, and restore outcome remain replayable, so destination can prove what resumed versus what was recreated locally. Design-partner access is available. Interested in one Go handoff experiment? If not, reply “pass” and I will not follow up.

PR #152, opened 2026-09-06
feat: add durable machine handoff ownership protocol

0xkaushik-ai:
Machine handoff needs durable execution ownership before it can safely stop one agent and start another. This draft adds the internal ownership coordinator and SQLite records: the source releases an activation token only after relinquishment commits, and a destination activation retry reconciles the recorded generation instead of launching again.

Part of #136. **This does not implement a working machine-to-machine session move and must remain draft.** There is no production runtime adapter, checkpoint exporter/importer, transport, HTTP/CLI/UI entry point, or consumer enforcing these reservations in the existing session manager yet. The runtime in the protocol tests is a fake; native conversation portability and real two-machine acceptance have not been verified. Source reservation must be integrated atomically with the existing mutation/input gates before this coordinator is exposed.

The protocol binds the handoff to a source, destination, session, checkpoint digest and activation verifier. Revision-based writes reject stale or rebound transitions. Relinquished source records remain durable tombstones; cancellation cannot reopen them. Unknown activation outcomes retain the destination reservation and require reconciliation. Ordinary service reads and CDC events omit the activation secret. Runtime operations and cancellation are serialized with context-aware admission.

Migration 126 avoids versions 119–125 already allocated in upstream intake #150; its landing order must be coordinated with that intake. SQL bindings are generated from source. No README, frontend, API contract, gateway policy or release changes.

Validation:

- Focused handoff service/store tests pass with `-race`, covering duplicate requests, concurrent activation, lost acknowledgments, failed persistence, unknown probes, cancellation, stale/rebound writes, CDC payloads and a real database close/reopen.
- `go test ./internal/storage/sqlite/... ./internal/service/handoff ./internal/domain` passes.
- Pinned golangci-lint v2.12.2 reports zero issues across the changed package families.
- `npm run sqlc` and `git diff --check` pass.
- Full `go test ./...` fails in `TestCopilotClassicPATIsNotAnAuthorizationSignal` and `TestFullLifecycleSpawnToTermination`. Both reproduce on an untouched archive of base `9b3b126c`: unexpected authorized Copilot status and missing fake-agent `events.log`, respectively.

Remaining for #136: select and verify the first native provider's portable state; implement secret-safe workspace/context checkpoints; enforce durable reservations throughout runtime startup, restore, input and lifecycle paths; implement authenticated transfer and client commands; and exercise crash/partition recovery with two actual machines. The activation grant must only travel over the authenticated transport, never in the checkpoint or ordinary progress response.


0xkaushik-ai:
@AgentWrapper

0xkaushik-ai:
@harshitsinghbhandari  do check this out

PR #154, opened 2026-09-07
feat: execute verified hao setup plans

### LINKED ISSUE #153: feat: execute verified hao setup plans
> *This was generated by AI during triage.*

## Goal

Make `hao setup` perform the trusted plan introduced in Batch 4, so a supported user-controlled machine can move from observed state to prepared AO host state safely and repeatably.

## Scope

- Resolve AO-owned artifacts only from authoritative immutable metadata containing version, source, and SHA-256 digest.
- Execute the same versioned action contract used by `hao setup --dry-run`.
- Support idempotent directory, file, package, artifact, service-definition, and Linux capability changes.
- Print the complete plan before mutation and require explicit confirmation.
- Separate unprivileged work from narrowly scoped privilege elevation.
- Verify downloads before replacement and use atomic writes, backups, and transaction journaling.
- On failure, report completed actions, rollback results, safe retry instructions, and stable JSON errors without secrets.
- Preserve persistent gateway identity, Hosted state roots, loopback daemon binding, and all gateway deny policies.
- Support clean Ubuntu x64 and arm64 first. Keep macOS local preparation non-privileged or manual where service management is not supported. Keep Windows remote service setup deferred.

## Explicitly out of scope

- `hao init`, pair certificate or passcode creation, service activation, and pairing-string output.
- Authentication management for coding agents or GitHub.
- Domain, ACME, Cloudflared, tunnel, relay, or public exposure work.
- Desktop publishing, signing, or changes to AO orchestration commands.
- Mutable `latest` installers, self-attested checksum sidecars, or unpinned shell installers.

## Acceptance criteria

- Dry-run and execution consume one shared action schema, with a drift test.
- Missing trusted immutable metadata remains deterministic `BLOCKED/manual`; it is never guessed.
- Re-running after success is a no-op.
- Interrupted and failed runs can explain, resume, or roll back without deleting pre-existing packages or AO data.
- Privileged commands are structured argv operations, never interpolated shell strings.
- Tests cover confirmation refusal, privilege refusal, digest mismatch, download interruption, concurrent invocation, rollback, crash recovery, hostile paths, redaction, and exact exit codes.
- Full backend race, lint, API drift, Linux x64/arm64 artifacts, and clean-machine packaged acceptance pass.
- PR receives independent architecture and security review before merge.

## Reference

`docs/hao-machine-management-boundary.md`, especially `hao setup`, update ownership, idempotency, privilege model, and error contract.

harshitsinghbhandari:
## Summary

- Execute the same validated setup action schema emitted by `hao setup --dry-run`.
- Require interactive typed consent or `--non-interactive --yes` before mutation.
- Resolve AO artifacts only from release-embedded version, source, and SHA-256 metadata.
- Apply idempotent directory, mode, artifact, package, vendor, and systemd definition actions.
- Verify bounded downloads before same-filesystem atomic replacement.
- Use structured privileged argv, durable journals, backups, rollback, crash recovery, and per-root locking.
- Preserve service activation, initialization, pairing, authentication, and public exposure as deferred work.

## Verification

- `go test ./internal/haocli -count=1 -timeout=90s`
- `go test -race -timeout=15m ./...` in a clean AO environment
- `go vet ./...`
- `golangci-lint v2.12.2 run ./internal/haocli`
- OpenAPI spec and frontend type regeneration with no drift
- Windows amd64 HAO package compilation
- Linux amd64 and arm64 AO plus HAO release-style builds with embedded artifact metadata

## Review

Independent architecture and security reviews requested against exact head `e419a5351`.

Closes #153

### DIFF
--- a/backend/internal/haocli/managed_open_windows.go
+++ b/backend/internal/haocli/managed_open_windows.go
@@ -169,7 +169,7 @@ func managedCreateExclusive(path string, mode os.FileMode) (*os.File, error) {
 }
 
 func managedCreateExclusiveAt(parent windows.Handle, name, path string, mode os.FileMode) (*os.File, error) {
-	handle, err := openManagedRelative(parent, name, windows.FILE_GENERIC_WRITE|windows.SYNCHRONIZE, windows.FILE_CREATE, windows.FILE_NON_DIRECTORY_FILE|windows.FILE_SYNCHRONOUS_IO_NONALERT)
+	handle, err := openManagedRelative(parent, name, windows.FILE_GENERIC_WRITE|windows.FILE_READ_ATTRIBUTES|windows.SYNCHRONIZE, windows.FILE_CREATE, windows.FILE_NON_DIRECTORY_FILE|windows.FILE_SYNCHRONOUS_IO_NONALERT)
 	if err != nil {
 		return nil, normalizeManagedWindowsError(err)
 	}
--- a/backend/internal/haocli/managed_open_windows.go
+++ b/backend/internal/haocli/managed_open_windows.go
@@ -12,29 +12,26 @@ import (
 	"golang.org/x/sys/windows"
 )
 
+const managedWindowsShareMode = windows.FILE_SHARE_READ | windows.FILE_SHARE_WRITE | windows.FILE_SHARE_DELETE
+
+type managedWindowsRenameInformation struct {
+	ReplaceIfExists uint32
+	RootDirectory   windows.Handle
+	FileNameLength  uint32
+	FileName        [1]uint16
+}
+
 func openManagedRegular(path string, maxSize int64) (*os.File, error) {
-	abs, err := filepath.Abs(filepath.Clean(path))
+	parent, name, err := openManagedParent(path)
 	if err != nil {
 		return nil, err
 	}
-	ntPath := `\??\` + abs
-	if strings.HasPrefix(abs, `\\`) {
-		ntPath = `\??\UNC\` + strings.TrimPrefix(abs, `\\`)
-	}
-	name, err := windows.NewNTUnicodeString(ntPath)
+	defer windows.CloseHandle(parent)
+	handle, err := openManagedRelative(parent, name, windows.FILE_GENERIC_READ|windows.SYNCHRONIZE, windows.FILE_OPEN, windows.FILE_NON_DIRECTORY_FILE|windows.FILE_SYNCHRONOUS_IO_NONALERT)
 	if err != nil {
-		return nil, err
-	}
-	attributes := &windows.OBJECT_ATTRIBUTES{ObjectName: name, Attributes: windows.OBJ_CASE_INSENSITIVE | windows.OBJ_DONT_REPARSE}
-	attributes.Length = uint32(unsafe.Sizeof(*attributes))
-	var handle windows.Handle
-	var status windows.IO_STATUS_BLOCK
-	if err := windows.NtCreateFile(&handle, windows.FILE_GENERIC_READ|windows.SYNCHRONIZE, attributes, &status, nil, 0,
-		windows.FILE_SHARE_READ|windows.FILE_SHARE_WRITE|windows.FILE_SHARE_DELETE, windows.FILE_OPEN,
-		windows.FILE_NON_DIRECTORY_FILE|windows.FILE_SYNCHRONOUS_IO_NONALERT, 0, 0); err != nil {
 		return nil, errors.New("managed file could not be opened without traversing a reparse point")
 	}
-	file := os.NewFile(uintptr(handle), abs)
+	file := os.NewFile(uintptr(handle), path)
 	info, err := file.Stat()
 	if err != nil || !info.Mode().IsRegular() || info.Size() > maxSize {
 		_ = file.Close()
@@ -43,47 +40,214 @@ func openManagedRegular(path string, maxSize int64) (*os.File, error) {
 	return file, nil
 }
 
+func openManagedParent(path string) (windows.Handle, string, error) {
+	abs, err := filepath.Abs(filepath.Clean(path))
+	if err != nil {
+		return 0, "", err
+	}
+	name := filepath.Base(abs)
+	parentPath := filepath.Dir(abs)
+	if name == "." || name == string(filepath.Separator) || parentPath == abs {
+		return 0, "", errors.New("managed path has no file component")
+	}
+	handle, err := openManagedAbsolute(parentPath, windows.FILE_LIST_DIRECTORY|windows.FILE_TRAVERSE|windows.FILE_READ_ATTRIBUTES|windows.SYNCHRONIZE, windows.FILE_DIRECTORY_FILE|windows.FILE_SYNCHRONOUS_IO_NONALERT)
+	if err != nil {
+		if isManagedWindowsNotExist(err) {
+			return 0, "", os.ErrNotExist
+		}
+		return 0, "", errors.New("managed path contains an inaccessible or reparse ancestor")
+	}
+	return handle, name, nil
+}
+
+func openManagedAbsolute(path string, access uint32, options uint32) (windows.Handle, error) {
+	ntPath := `\??\` + path
+	if strings.HasPrefix(path, `\\`) {
+		ntPath = `\??\UNC\` + strings.TrimPrefix(path, `\\`)
+	}
+	name, err := windows.NewNTUnicodeString(ntPath)
+	if err != nil {
+		return 0, err
+	}
+	attributes := managedWindowsObjectAttributes(0, name)
+	return ntCreateManagedWindowsFile(attributes, access, windows.FILE_OPEN, options)
+}
+
+func openManagedRelative(parent windows.Handle, name string, access uint32, disposition uint32, options uint32) (windows.Handle, error) {
+	if name == "" || name == "." || name == ".." || strings.ContainsAny(name, `/\`) {
+		return 0, errors.New("managed path has invalid file component")
+	}
+	objectName, err := windows.NewNTUnicodeString(name)
+	if err != nil {
+		return 0, err
+	}
+	attributes := managedWindowsObjectAttributes(parent, objectName)
+	return ntCreateManagedWindowsFile(attributes, access, disposition, options)
+}
+
+func managedWindowsObjectAttributes(parent windows.Handle, name *windows.NTUnicodeString) *windows.OBJECT_ATTRIBUTES {
+	attributes := &windows.OBJECT_ATTRIBUTES{
+		RootDirectory: parent,
+		ObjectName:    name,
+		Attributes:    windows.OBJ_CASE_INSENSITIVE | windows.OBJ_DONT_REPARSE,
+	}
+	attributes.Length = uint32(unsafe.Sizeof(*attributes))
+	return attributes
+}
+
+func ntCreateManagedWindowsFile(attributes *windows.OBJECT_ATTRIBUTES, access uint32, disposition uint32, options uint32) (windows.Handle, error) {
+	var handle windows.Handle
+	var status windows.IO_STATUS_BLOCK
+	err := windows.NtCreateFile(&handle, access, attributes, &status, nil, windows.FILE_ATTRIBUTE_NORMAL,
+		managedWindowsShareMode, disposition, options, 0, 0)
+	return handle, err
+}
+
 func managedLstat(path string) (os.FileInfo, error) {
-	if err := ensureNoSymlinkAncestors(path); err != nil {
+	parent, name, err := openManagedParent(path)
+	if err != nil {
 		return nil, err
 	}
-	return os.Lstat(path)
+	defer windows.CloseHandle(parent)
+	handle, err := openManagedRelative(parent, name, windows.FILE_READ_ATTRIBUTES|windows.SYNCHRONIZE, windows.FILE_OPEN, windows.FILE_SYNCHRONOUS_IO_NONALERT)
+	if err != nil {
+		if isManagedWindowsNotExist(err) {
+			return nil, os.ErrNotExist
+		}
+		return nil, errors.New("managed target is inaccessible or a reparse point")
+	}
+	file := os.NewFile(uintptr(handle), path)
+	defer file.Close()
+	return file.Stat()
 }
 
 func managedChmod(path string, mode os.FileMode) error {
-	if _, err := managedLstat(path); err != nil {
+	parent, name, err := openManagedParent(path)
+	if err != nil {
 		return err
 	}
-	return os.Chmod(path, mode)
+	defer windows.CloseHandle(parent)
+	return managedChmodAt(parent, name, path, mode)
+}
+
+func managedChmodAt(parent windows.Handle, name, path string, mode os.FileMode) error {
+	handle, err := openManagedRelative(parent, name, windows.FILE_READ_ATTRIBUTES|windows.FILE_WRITE_ATTRIBUTES|windows.SYNCHRONIZE, windows.FILE_OPEN, windows.FILE_SYNCHRONOUS_IO_NONALERT)
+	if err != nil {
+		return normalizeManagedWindowsError(err)
+	}
+	file := os.NewFile(uintptr(handle), path)
+	defer file.Close()
+	return file.Chmod(mode)
 }
 
 func managedMkdir(path string, mode os.FileMode) error {
-	if err := ensureNoSymlinkAncestors(filepath.Dir(path)); err != nil {
+	parent, name, err := openManagedParent(path)
+	if err != nil {
 		return err
 	}
-	return os.Mkdir(path, mode)
+	defer windows.CloseHandle(parent)
+	return managedMkdirAt(parent, name, path, mode)
+}
+
+func managedMkdirAt(parent windows.Handle, name, path string, mode os.FileMode) error {
+	handle, err := openManagedRelative(parent, name, windows.FILE_LIST_DIRECTORY|windows.FILE_TRAVERSE|windows.FILE_READ_ATTRIBUTES|windows.FILE_WRITE_ATTRIBUTES|windows.SYNCHRONIZE, windows.FILE_CREATE, windows.FILE_DIRECTORY_FILE|windows.FILE_SYNCHRONOUS_IO_NONALERT)
+	if err != nil {
+		return normalizeManagedWindowsError(err)
+	}
+	file := os.NewFile(uintptr(handle), path)
+	defer file.Close()
+	return file.Chmod(mode)
 }
 
 func managedCreateExclusive(path string, mode os.FileMode) (*os.File, error) {
-	if err := ensureNoSymlinkAncestors(filepath.Dir(path)); err != nil {
+	parent, name, err := openManagedParent(path)
+	if err != nil {
 		return nil, err
 	}
-	return os.OpenFile(path, os.O_CREATE|os.O_EXCL|os.O_WRONLY, mode)
+	defer windows.CloseHandle(parent)
+	return managedCreateExclusiveAt(parent, name, path, mode)
+}
+
+func managedCreateExclusiveAt(parent windows.Handle, name, path string, mode os.FileMode) (*os.File, error) {
+	handle, err := openManagedRelative(parent, name, windows.FILE_GENERIC_WRITE|windows.SYNCHRONIZE, windows.FILE_CREATE, windows.FILE_NON_DIRECTORY_FILE|windows.FILE_SYNCHRONOUS_IO_NONALERT)
+	if err != nil {
+		return nil, normalizeManagedWindowsError(err)
+	}
+	file := os.NewFile(uintptr(handle), path)
+	if err := file.Chmod(mode); err != nil {
+		_ = file.Close()
+		return nil, err
+	}
+	return file, nil
 }
 
 func managedRename(source, destination string) error {
-	if err := ensureNoSymlinkAncestors(filepath.Dir(source)); err != nil {
+	sourceParent, sourceName, err := openManagedParent(source)
+	if err != nil {
 		return err
 	}
-	if err := ensureNoSymlinkAncestors(filepath.Dir(destination)); err != nil {
+	defer windows.CloseHandle(sourceParent)
+	destinationParent, destinationName, err := openManagedParent(destination)
+	if err != nil {
 		return err
 	}
-	return os.Rename(source, destination)
+	defer windows.CloseHandle(destinationParent)
+	sourceHandle, err := openManagedRelative(sourceParent, sourceName, windows.DELETE|windows.SYNCHRONIZE, windows.FILE_OPEN, windows.FILE_SYNCHRONOUS_IO_NONALERT)
+	if err != nil {
+		return normalizeManagedWindowsError(err)
+	}
+	defer windows.CloseHandle(sourceHandle)
+	return renameManagedWindowsHandle(sourceHandle, destinationParent, destinationName)
+}
+
+func renameManagedWindowsHandle(source, destinationParent windows.Handle, destinationName string) error {
+	name, err := windows.UTF16FromString(destinationName)
+	if err != nil {
+		return err
+	}
+	nameLength := (len(name) - 1) * 2
+	var layout managedWindowsRenameInformation
+	bufferSize := int(unsafe.Offsetof(layout.FileName)) + nameLength
+	buffer := make([]byte, bufferSize)
+	info := (*managedWindowsRenameInformation)(unsafe.Pointer(&buffer[0]))
+	info.ReplaceIfExists = windows.FILE_RENAME_REPLACE_IF_EXISTS
+	info.RootDirectory = destinationParent
+	info.FileNameLength = uint32(nameLength)
+	copy((*[windows.MAX_LONG_PATH]uint16)(unsafe.Pointer(&info.FileName[0]))[:nameLength/2:nameLength/2], name)
+	var status windows.IO_STATUS_BLOCK
+	return windows.NtSetInformationFile(source, &status, &buffer[0], uint32(bufferSize), windows.FileRenameInformation)
 }
 
 func managedRemove(path string) error {
-	if err := ensureNoSymlinkAncestors(filepath.Dir(path)); err != nil {
+	parent, name, err := openManagedParent(path)
+	if err != nil {
 		return err
 	}
-	return os.Remove(path)
+	defer windows.CloseHandle(parent)
+	return managedRemoveAt(parent, name)
+}
+
+func managedRemoveAt(parent windows.Handle, name string) error {
+	handle, err := openManagedRelative(parent, name, windows.DELETE|windows.SYNCHRONIZE, windows.FILE_OPEN, windows.FILE_SYNCHRONOUS_IO_NONALERT)
+	if err != nil {
+		return normalizeManagedWindowsError(err)
+	}
+	defer windows.CloseHandle(handle)
+	deleteFile := byte(1)
+	var status windows.IO_STATUS_BLOCK
+	return windows.NtSetInformationFile(handle, &status, &deleteFile, 1, windows.FileDispositionInformation)
+}
+
+func isManagedWindowsNotExist(err error) bool {
+	return errors.Is(err, windows.STATUS_OBJECT_NAME_NOT_FOUND) || errors.Is(err, windows.STATUS_OBJECT_PATH_NOT_FOUND)
+}
+
+func normalizeManagedWindowsError(err error) error {
+	if isManagedWindowsNotExist(err) {
+		return os.ErrNotExist
+	}
+	if errors.Is(err, windows.STATUS_OBJECT_NAME_COLLISION) {
+		return os.ErrExist
+	}
+	return err
 }
--- a/backend/internal/haocli/managed_open_windows.go
+++ b/backend/internal/haocli/managed_open_windows.go
@@ -42,3 +42,48 @@ func openManagedRegular(path string, maxSize int64) (*os.File, error) {
 	}
 	return file, nil
 }
+
+func managedLstat(path string) (os.FileInfo, error) {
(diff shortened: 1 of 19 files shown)


backend/internal/haocli/setup_execute.go
```diff
@@ -0,0 +1,914 @@
...
+	Run(context.Context, bool, bool, io.Reader, string, ...string) error
+}
+
+type systemSetupExecution struct{}
+
+func (systemSetupExecution) Download(ctx context.Context, source string) (io.ReadCloser, error) {
+	if !strings.HasPrefix(source, "https://") {
+		return nil, errors.New("artifact source is not HTTPS")
+	}
+	client := &http.Client{Timeout: 10 * time.Minute, CheckRedirect: func(*http.Request, []*http.Request) error { return http.ErrUseLastResponse }}
```
harshitsinghbhandari: [P1] Allow only the expected GitHub release redirect. Every trusted source emitted by `trustedBuildArtifact` is a `github.com/.../releases/download/...` URL, and GitHub answers those requests with HTTP 302 to its release asset host. Because this client rejects every redirect, a real probe against the published `v0.13.1/ao-linux-x64` asset fails here with `artifact download returned HTTP 302`, so production `hao setup` cannot install the embedded artifact. Follow a bounded number of HTTPS redirects only to an explicit GitHub release asset host allowlist, then keep the existing size and digest checks on the final body.

backend/internal/haocli/setup_execute.go
```diff
@@ -0,0 +1,914 @@
...
+		stateRoot:   filepath.Clean(stateRoot),
+		journalPath: filepath.Join(stateRoot, ".hao-setup-transaction.json"),
+		options:     options,
+		result: SetupExecutionResult{
+			Completed: []string{},
+			Rollback:  []string{},
+			Recovery:  []string{},
+		},
+	}
+	if recovered, recoverErr := e.recover(ctx); recoverErr != nil {
```
harshitsinghbhandari: [P1] Preflight and disclose privilege needed by recovery. `recover` runs before `planNeedsPrivilege` and may call privileged `rm`, `cp`, or `mv` for an interrupted service record. A new plan with no privileged actions can therefore invoke sudo that was absent from the displayed plan; refusal is then wrapped as `setup_failed` with exit 1 instead of the required `privilege_required` exit 3. I reproduced this with a valid created-service journal and an empty current plan. Inspect the validated journal for privileged recovery first, include that recovery in consent, and run `CheckPrivilege` before rollback.

backend/internal/haocli/setup_execute.go
```diff
@@ -0,0 +1,914 @@
...
+	}
+	return e.completeLastRecord()
+}
+
+func (e *setupExecutor) runIrreversible(ctx context.Context, stepID, kind string, privileged bool, executable string, argv []string) error {
+	record := journalRecord{StepID: stepID, Kind: kind, Reversible: false}
+	if err := e.record(record); err != nil {
+		return err
+	}
+	if err := e.system.Run(ctx, privileged, e.options.NonInteractive, e.options.Input, executable, argv...); err != nil {
```
harshitsinghbhandari: [P1] Bind package execution to the declared source and digest. `SetupPackageAction` and `SetupVendorAction` carry immutable `Source` and `SHA256` fields, but this path ignores both and asks apt or npm to resolve the version again. The executed bytes can differ from the consented and journaled tuple while the plan still appears cryptographically pinned. Either download and verify the declared bytes before a fixed local install argv, or remove these action kinds until a real package-manager integrity contract is implemented.

backend/internal/haocli/setup_execute.go
```diff
@@ -0,0 +1,914 @@
...
+		return err
+	}
+	temporary, err := randomTemporaryPath(filepath.Dir(action.Path), ".hao-artifact-")
+	if err != nil {
+		return err
+	}
+	record := journalRecord{StepID: stepID, Kind: "verified-release-artifact", Target: action.Path, ManifestTarget: action.Path + ".hao-manifest.json", Temporary: temporary, Reversible: true}
+	if targetErr == nil {
+		record.Backup = filepath.Join(e.backupDir, safeStepName(stepID)+".binary")
+		if err := copyRegularFile(action.Path, record.Backup, 0o700); err != nil {
```
harshitsinghbhandari: [P2] Preserve artifact metadata across rollback. This backup is always written as mode 0700 and `rollbackManagedFile` also restores it as 0700, regardless of the pre-run mode. A reproduced update from a valid 0755 AO binary followed by a later action failure restores the old bytes but leaves the binary at 0700. Record the original mode and restore it, as the file-mode journal path already does, so rollback actually returns the pre-run file state.

backend/internal/haocli/setup_execute.go
```diff
@@ -0,0 +1,914 @@
...
+		return e.runIrreversible(ctx, step.ID, a.Kind, true, a.Package.Executable, a.Package.Argv)
+	case "vendor-package":
+		return e.runIrreversible(ctx, step.ID, a.Kind, false, a.Vendor.Executable, a.Vendor.Argv)
+	default:
+		return fmt.Errorf("unsupported action kind %q", a.Kind)
+	}
+}
+
+func (e *setupExecutor) createDirectory(stepID string, action *SetupFileAction) error {
+	if err := ensureNoSymlinkAncestors(filepath.Dir(action.Path)); err != nil {
```
harshitsinghbhandari, line 394: [P2] Make mutations relative to no-follow directory handles. This check and the later `Lstat`, `Mkdir`, `Chmod`, `CreateTemp`, and `Rename` calls are separate pathname resolutions. A concurrent process can replace a checked state-root ancestor with a symlink between them and redirect a managed write outside the Hosted AO root. The read path already uses `openat` plus `O_NOFOLLOW`; the write path needs the same handle-relative walk and identity checks, with the Windows equivalent rejecting reparse traversal.

harshitsinghbhandari (commented):
True verdict: REQUEST_CHANGES. Submitted as COMMENTED only because GitHub prohibits requesting changes on one's own pull request.

The artifact tuple is correctly bound for Linux x64 and arm64, and the narrow race suite passes, but the production install path is not usable and three safety contracts are violated. Aggressive boundary verification failed with 10 unverified boundaries out of 19 total. The inline findings are blocking.

Verification: `go test -race ./internal/haocli -count=1 -timeout=3m`, `go vet ./internal/haocli`, release-style Linux x64 and arm64 builds with real `trustedBuildArtifact` consumption, a real published-asset download probe, interrupted privilege recovery reproduction, and rollback mode reproduction.

backend/internal/haocli/setup_execute.go
```diff
@@ -0,0 +1,914 @@
...
+
+func writeArtifactManifest(action *SetupArtifactAction) error {
+	data, err := json.Marshal(ArtifactMetadata{Version: action.Version, Source: action.Source, SHA256: action.SHA256})
+	if err != nil {
+		return err
+	}
+	return atomicWriteFile(action.Path+".hao-manifest.json", append(data, '\n'), 0o600)
+}
+
+func (e *setupExecutor) installService(ctx context.Context, stepID, disposition string, action *SetupServiceAction) error {
```
harshitsinghbhandari, line 590: [P1] Preserve the resolved AO data and run-file paths in installed units. This now installs definitions rendered only from `StateRoot`. `renderSystemdDefinition` reconstructs `<root>/data` and `<root>/running.json`, even though `ResolveStateRoot` deliberately collapses independent `AO_DATA_DIR` and `AO_RUN_FILE` overrides. The gateway definition exports neither variable, so `ao vm serve` can also load defaults and miss the daemon. A regression probe with both overrides under different parents produced values that matched neither configured path. Carry the exact resolved data directory and run file into the desired state and render both services from them.

backend/internal/haocli/setup_execute.go
```diff
@@ -0,0 +1,914 @@
...
+}
+
+func (e *setupExecutor) installArtifact(ctx context.Context, stepID, disposition string, action *SetupArtifactAction) error {
+	if err := ensureNoSymlinkAncestors(action.Path); err != nil {
+		return err
+	}
+	if digest, err := fileDigest(action.Path, maxArtifactSize); err == nil && digest == action.SHA256 {
+		return writeArtifactManifest(action)
+	}
+	_, targetErr := os.Lstat(action.Path)
```
harshitsinghbhandari: [P1] Revalidate the update precondition before replacement. For an `update`, this only rejects disappearance. If another process changes the artifact after the plan is printed and approved, the executor backs up that new file and overwrites it with the stale plan. A probe reproduced that overwrite. Put the observed digest or identity in the action and compare it immediately before mutation. `installService` has the same gap for changed service content, so both update paths need the precondition.

harshitsinghbhandari (commented):
## PR Review: #154

### Summary

The release tuple and production planner correctly keep package and vendor installation blocked without a trusted resolver, and the focused tests, race suite, release-style metadata build, and all 14 CI checks pass. Four correctness failures remain in state propagation, stale-plan safety, journal recovery, and JSON output.

### Boundary Verification

**FAIL:** 6 of 18 traced boundaries are unverified and reproduced as failing joins. Twelve are verified, with no deferrals or overrides. The failed seams are explicit state overrides into daemon and gateway units, stale artifact and service updates, service backup records into recovery, and execution results into a JSON consumer.

### Recommendation

**REQUEST CHANGES.** This review is submitted as COMMENTED because the authenticated reviewer is the PR author, but the true verdict is REQUEST CHANGES.

BOUNDARY GATE FAILED: 6 unverified boundaries, see the review findings.

backend/internal/haocli/setup_execute.go
```diff
@@ -0,0 +1,914 @@
...
+	record := journalRecord{StepID: stepID, Kind: "service-definition-v1", Target: action.Path, Temporary: targetTemp, Reversible: true}
+	if targetErr == nil {
+		record.Backup, err = randomTemporaryPath(filepath.Dir(action.Path), ".hao-backup-")
+		if err != nil {
+			return err
+		}
+	} else {
+		record.Created = true
+	}
+	if err := e.record(record); err != nil {
```
harshitsinghbhandari, line 627: [P1] Do not persist a service restore record before its backup exists. The journal is written here, then the privileged copy runs. If that copy fails before any target mutation, rollback tries to restore the nonexistent backup, reports incomplete recovery, and every later run retries the same impossible restore. A regression probe reproduced the poisoned journal. Create and durably finish the backup before recording a state that requires it, or represent the prepared state so recovery can safely discard it.

backend/internal/haocli/setup.go
```diff
@@ -181,15 +181,78 @@ func newSetupCommand(deps Deps, opts *options) *cobra.Command {
...
+		}
+		result, executeErr := deps.ExecuteSetup(cmd.Context(), plan, desired.StateRoot, SetupExecutionOptions{NonInteractive: nonInteractive, Input: cmd.InOrStdin()})
+		if executeErr != nil {
+			if !opts.json {
+				writeSetupFailure(cmd.OutOrStdout(), result)
+			}
+			return executeErr
+		}
+		if opts.json {
+			return writeJSON(cmd.OutOrStdout(), haocontractRedact(result))
```
harshitsinghbhandari: [P2] Emit one JSON document for a successful execution. The command already writes the plan before consent, then this writes a second top-level object. `json.Unmarshal` of successful `hao --json setup --non-interactive --yes` output fails with `invalid character '{' after top-level value`. Return one versioned envelope containing the plan and result, and make the execution test decode the complete output instead of ignoring `execOut`.

backend/internal/haocli/setup_execute.go
harshitsinghbhandari: Fixed in 2a2de42df. Service records are journaled before backup creation and gain a durable backupReady marker only after copy success. Recovery removes incomplete backups and never restores them.

backend/internal/haocli/setup.go
harshitsinghbhandari: Fixed in 2a2de42df. JSON execution now emits one schemaVersion 1 document containing both plan and result, including refusal and failure results. The drift test decodes the complete execution report.

backend/internal/haocli/setup_execute.go
harshitsinghbhandari: Fixed in 2a2de42df. Desired state resolves the exact AO_DATA_DIR and AO_RUN_FILE values and renders both into daemon and gateway units. A regression test covers independent overrides.

backend/internal/haocli/setup_execute.go
harshitsinghbhandari: Fixed in 2a2de42df. Package and vendor actions now download the authoritative Source, verify SHA-256, and substitute only the verified local file into the allowlisted structured argv operation.

backend/internal/haocli/setup_execute.go
harshitsinghbhandari: Fixed in 2a2de42df. Recovery is now a displayed versioned plan action bound to the journal SHA-256. Its required privilege is derived from the validated journal, checked before recovery, and refusal returns exit 3 without mutation.

backend/internal/haocli/setup_execute.go
harshitsinghbhandari: Fixed in 2a2de42df. Managed mutations now use no-follow parent directory handles and *at operations on Unix. Windows checks reparse ancestors. Atomic writes reject linked targets, with hostile target and ancestor tests.

backend/internal/haocli/setup_execute.go
harshitsinghbhandari: Fixed in 2a2de42df. Update actions carry observed SHA-256 preconditions, plus artifact mode, and execution rejects changed or disappeared targets before replacement. Stale artifact execution is covered.

backend/internal/haocli/setup_execute.go
harshitsinghbhandari: Fixed in 2a2de42df. Downloads now allow only bounded HTTPS redirects from the immutable agentlab-in/hosted-ao release path to the exact release-assets.githubusercontent.com host, with allowlist tests for scheme and host confusion.

backend/internal/haocli/setup_execute.go
harshitsinghbhandari: Fixed in 2a2de42df. Artifact journal records preserve the original executable mode and rollback restores that exact mode. A regression test covers mode restoration.

backend/internal/haocli/setup_execute.go
```diff
@@ -0,0 +1,1107 @@
...
+	return e.completeLastRecord()
+}
+
+func (e *setupExecutor) installArtifact(ctx context.Context, stepID, disposition string, action *SetupArtifactAction) error {
+	if err := ensureNoSymlinkAncestors(action.Path); err != nil {
+		return err
+	}
+	currentDigest, digestErr := fileDigest(action.Path, maxArtifactSize)
+	binaryMatches := digestErr == nil && currentDigest == action.SHA256
+	if binaryMatches && artifactManifestMatches(action) {
```
harshitsinghbhandari: [P1] Validate the consented precondition before this desired-state fast path. I reproduced an update planned from an observed mode 0700 artifact, then replaced it after consent with the desired bytes at mode 0666. Because `binaryMatches` is true, execution skips both `ExpectedSHA256` and `ExpectedMode`, writes the trusted manifest, reports `completed`, and leaves the trusted executable group and world writable. If the manifest already matches, this line returns even earlier. `installService` has the parallel issue at its digest fast path: desired content returns before the expected digest and file-safety checks, so a mode 0666 definition is accepted as complete. Validate the observed digest, recorded mode where applicable, and current file safety before treating convergence as success.

harshitsinghbhandari (commented):
Verdict: REQUEST CHANGES

Aggressive architecture boundary gate: 16 verified, 2 unverified, 0 deferred. Eight of the nine prior thread fixes now hold under focused verification. The stale-plan fix remains incomplete at the already-converged digest boundary, affecting both artifact and service updates.

Verified: exact state overrides reach both systemd units; recovery is consented, digest-bound, and privilege-preflighted; incomplete service backups recover safely; successful JSON execution is one document; rollback restores the artifact mode; package and vendor execution consumes only verified local files while production planning remains BLOCKED without trusted metadata; bounded real GitHub release redirect succeeds; managed Unix writes use no-follow handles and Windows amd64 compiles. The focused haocli package passed, the published v0.13.1 artifact passed the production downloader, and all 14 CI checks were green at the final inspection.

No generic vendor shell executor was introduced. No scope expansion is requested.

backend/internal/haocli/setup.go
```diff
@@ -206,12 +327,23 @@ func resolveSetupDesired(deps Deps, path string, object map[string]any, install
 		return setupDesired{}, operationalError("resolve state root", err)
 	}
 	d.StateRoot = root
+	d.DataDir, err = deps.DataDir()
```
harshitsinghbhandari, line 334: [P1] Prepare the resolved data directory that the unit consumes. This stores the exact AO_DATA_DIR override in desired state, but observeSetup still hardcodes directory.data to filepath.Join(StateRoot, "data") and the executor allowlist repeats that path. With AO_DATA_DIR=<stateRoot>/custom-db on a clean host, the accepted plan creates <stateRoot>/data while both units use custom-db as WorkingDirectory and AO_DATA_DIR. A focused planner-to-unit repro fails on that mismatch, so setup can report completion while the installed daemon and gateway units cannot start because their working directory was never created. It can also manage unrelated sibling paths when ResolveStateRoot derives the parent of a nonstandard override. Thread DataDir through the directory observation, plan action, and executor target validation.

backend/internal/haocli/managed_open_windows.go
```diff
@@ -42,3 +42,48 @@ func openManagedRegular(path string, maxSize int64) (*os.File, error) {
 	}
 	return file, nil
 }
+
+func managedLstat(path string) (os.FileInfo, error) {
+	if err := ensureNoSymlinkAncestors(path); err != nil {
```
harshitsinghbhandari: [P1] Keep the Windows mutation on a no-reparse handle. Each new write helper checks ancestors here and then calls os.Chmod, os.Mkdir, os.OpenFile, os.Rename, or os.Remove by path in a later operation. A concurrent rename plus junction swap between those calls redirects the mutation outside the managed tree. The Windows tests only exercise openManagedRegular, whose NtCreateFile handle is safe, and do not exercise these mutation wrappers. Use handle-relative operations with reparse refusal, or open and validate a stable parent handle and mutate relative to it, matching the Unix fix.

harshitsinghbhandari (commented):
True verdict: REQUEST_CHANGES. GitHub rejects a request-changes review on this account's own pull request, so this is submitted as COMMENTED. Aggressive boundary gate: 17 verified, 5 unverified. The amended head fixes the prior redirect, digest, recovery privilege, rollback mode, backup readiness, JSON framing, and Unix no-follow cases under focused tests, including a live GitHub release redirect. Two actionable seams remain inline. The other unverified seams require real sudo/systemd, apt/npm, or as-yet unpublished x64/arm64 release artifacts and were not exercised on this host. GitHub showed 14 of 14 checks successful.

backend/internal/haocli/setup_execute.go
harshitsinghbhandari: Fixed in 81a0749b04400035976fe4cb8727b130b905d914. Artifact and service updates now validate the consented expected digest and expected mode before any desired-state fast path. Both also validate current file safety before success. Regressions cover desired artifact bytes at mode 0666, desired service content at mode 0666, and safe service mode drift, with no mutation or trusted manifest write.

backend/internal/haocli/setup.go
harshitsinghbhandari: Fixed in 81a0749b04400035976fe4cb8727b130b905d914. The exact resolved AO_DATA_DIR now flows through observation, the directory.data action, executable-plan validation, execution, journal recovery validation, and both service consumers. The planner-to-executor regression verifies the custom directory is created at mode 0700 and the default stateRoot/data path is not created.

backend/internal/haocli/managed_open_windows.go
harshitsinghbhandari: Fixed in 81a0749b04400035976fe4cb8727b130b905d914. Windows managed mutations now open a no-reparse parent handle once and perform create, mkdir, chmod, rename, and remove relative to that stable handle. The Windows regression swaps the visible parent to a junction after handle acquisition and verifies every mutation remains in the originally opened directory.

backend/internal/haocli/managed_open_windows_test.go
```diff
@@ -98,3 +100,162 @@ func TestOpenManagedRegularRejectsWindowsAncestorSwap(t *testing.T) {
 		}
 	}
 }
+
+func TestManagedWindowsMutationsUseStableParentHandles(t *testing.T) {
```
harshitsinghbhandari, line 109: This is the correct regression for the prior Windows TOCTOU defect, but it is never executed by the PR CI. The Go Windows job at .github/workflows/go.yml:110-129 only runs workspace and session-manager packages, and the Windows CLI E2E job only runs internal/cli. A GOOS=windows test build proves compilation, not the create, mkdir, chmod, remove, and rename behavior exercised here. Add this package or these named tests to a windows-latest job so the held-handle boundary is actually merge-gated.

harshitsinghbhandari (commented):
True verdict: REQUEST_CHANGES. GitHub prohibits requesting changes on this account own pull request, so this is submitted as COMMENTED. Re-review at exact 81a0749b04400035976fe4cb8727b130b905d914 closes the custom AO_DATA_DIR defect: the real planner, executor, recovery validator, daemon unit, and gateway unit now consume the same custom path, and the focused test plus race run pass. The Windows mutation implementation is structurally corrected to operate through stable no-reparse parent handles, and Windows amd64 compilation passes. Aggressive boundary gate: 1 verified, 1 unverified. The Windows-only hostile swap regression is not executed by any CI job, so the security boundary remains unverified until that test runs on Windows.

harshitsinghbhandari (commented):
Verdict: REQUEST CHANGES

The artifact and service stale/converged P1 is closed on this head. Independent probes confirm both desired-state fast paths now validate the consented digest and mode first and reject mode 0666 with zero mutation. The four original architecture repros are otherwise closed: exact data/run paths flow through planning and execution, incomplete service backups recover safely, and JSON execution emits one report.

One architecture P1 remains: the generated pair unit overrides the gateway's persistent certificate and passcode defaults with paths derived from the movable runtime root. This violates the accepted compatibility exception and can invalidate existing fingerprint pins.

Aggressive boundary gate: 17 verified, 2 unverified, 19 total, 0 deferred. The second unverified boundary is the changed Windows native mutation seam, whose Windows-only runtime tests are not exercised by PR CI; existing comment 3952229743 already covers that gap, so I did not duplicate it inline. Focused haocli and vmgateway packages pass, Windows amd64 compilation passes, and all 14 checks were green during inspection.

backend/internal/haocli/setup.go
```diff
@@ -589,15 +753,15 @@ func renderSystemdDefinition(id string, d setupDesired, user UserObservation) st
 		b.WriteString("daemon\nAfter=network-online.target\n")
 	}
 	path := filepath.Join(user.Home, ".local", "bin") + ":" + filepath.Join(user.Home, "bin") + ":/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
-	b.WriteString("\n[Service]\nType=simple\nUser=" + user.Name + "\nWorkingDirectory=" + dataDir + "\nEnvironment=" + systemdQuote("HOME="+user.Home) + "\nEnvironment=" + systemdQuote("PATH="+path) + "\n")
+	b.WriteString("\n[Service]\nType=simple\nUser=" + user.Name + "\nWorkingDirectory=" + dataDir + "\nEnvironment=" + systemdQuote("HOME="+user.Home) + "\nEnvironment=" + systemdQuote("PATH="+path) + "\nEnvironment=" + systemdQuote("AO_DATA_DIR="+dataDir) + "\nEnvironment=" + systemdQuote("AO_RUN_FILE="+runFile) + "\n")
 	if id == "service.gateway" {
 		b.WriteString("Environment=" + systemdQuote("AO_VM_PAIR=on") + "\nEnvironment=" + systemdQuote("AO_VM_HTTPS_ADDR=:"+strconv.Itoa(d.PairPort)) + "\nEnvironment=" + systemdQuote("AO_VM_CERT_DIR="+filepath.Join(d.StateRoot, "vm-gateway", "pair-cert")) + "\nEnvironment=" + systemdQuote("AO_VM_PASSCODE_DIR="+filepath.Join(d.StateRoot, "vm-gateway", "pair-passcode")) + "\n")
```
harshitsinghbhandari: [P1] Preserve the persistent pair identity root when runtime state moves. `d.StateRoot` follows `AO_RUN_FILE` and `AO_DATA_DIR`, but the accepted compatibility contract keeps the pair certificate and passcode fallbacks under `config.DefaultStateDir()` unless their own explicit gateway overrides are set. These exported values are treated as explicit by `vmgateway.Resolve`, so they bypass its correct defaults. I reproduced the real seam by rendering a gateway unit with a separate runtime root and passing its environment into `vmgateway.Resolve`: both paths resolved below `runtime/vm-gateway` instead of `~/.ao/hosted/vm-gateway`. An existing paired box can therefore start with a fresh certificate and passcode, invalidating fingerprint pins. Resolve identity-specific paths using `AO_VM_CERT_DIR` and `AO_VM_PASSCODE_DIR` precedence with `DefaultStateDir()` fallbacks, then round-trip the generated unit through the real resolver under data-only, run-only, and combined runtime overrides.

backend/internal/haocli/managed_open_windows_test.go
harshitsinghbhandari: Fixed in 71ce14c465ac1e9c872146ae0f6eac68e9060fc3. The existing windows-latest Go job now runs the Windows HAO managed-path tests under the race detector, including the hostile parent swap coverage for create, mkdir, chmod, remove, and rename.

backend/internal/haocli/setup.go
harshitsinghbhandari: Fixed in 71ce14c465ac1e9c872146ae0f6eac68e9060fc3. HAO resolves AO_VM_CERT_DIR and AO_VM_PASSCODE_DIR first, then uses canonical DefaultStateDir-backed pair identity helpers. The unit receives those exact identity paths instead of deriving them from the movable runtime root. A real resolver round-trip covers data-only, run-only, combined runtime overrides, and explicit identity overrides.

harshitsinghbhandari (commented):
HISTORICAL REVIEW OF 71ce14c465ac1e9c872146ae0f6eac68e9060fc3. This was posted after the PR had advanced to eaac2b6670ca0f597818606f363138b35df74251, so it is stale and is not a verdict on the current PR head. It was not an orchestrator-authorized current-head verdict. The findings are retained below as historical evidence. The superseding current-head architecture review is https://github.com/agentlab-in/hosted-ao/pull/154#pullrequestreview-5135056522, with true verdict APPROVE.

Historical verdict for 71ce14c465ac1e9c872146ae0f6eac68e9060fc3: REQUEST CHANGES

[P1] Windows setup cannot create managed temporary files. In backend/internal/haocli/managed_open_windows.go:172, managedCreateExclusiveAt opens the new file with FILE_GENERIC_WRITE | SYNCHRONIZE, then line 177 calls Chmod without requesting FILE_READ_ATTRIBUTES | FILE_WRITE_ATTRIBUTES. The real Windows check fails at managed_open_windows_test.go:109 with Access is denied. This production helper is used by artifact downloads at setup_execute.go:527 and atomic service and manifest writes at setup_execute.go:1020, so a Windows executor cannot apply either path. Request the attribute rights on the stable handle and retain a real Windows regression.

The persistent gateway identity issue is closed. I independently verified data-only, run-only, combined, and explicit AO_VM_CERT_DIR/AO_VM_PASSCODE_DIR cases through rendered systemd environment into the real vmgateway.Resolve consumer. The stale artifact and service fast paths, backup readiness, shared plan/result schema, rollback, provenance, and state-root checks remain closed on unchanged source.

Historical aggressive boundary gate: FAIL. 18 verified, 0 unverified, 1 failed, 19 total. CI evidence for 71ce14c465ac1e9c872146ae0f6eac68e9060fc3: Windows run 34157326600, job 101851751975.

harshitsinghbhandari (commented):
True verdict: APPROVE. Submitted as COMMENTED because GitHub prohibits approving this account own pull request.

The Windows production failure is closed. managedCreateExclusiveAt now supplies the read-attribute access required by Go handle-based Chmod, while generic write retains attribute-write access. The native Windows merge job passes the hostile ancestor swap and stable-parent create, directory, chmod, remove, and rename paths. The EOF and rename assertions now model Windows creation and handle lifetimes without weakening the no-escape invariant.

The persistent gateway identity correction is unchanged and remains verified through rendered systemd environment into real vmgateway.Resolve for data-only, run-only, combined, and explicit certificate/passcode overrides. Prior stale-plan, provenance, journal recovery, rollback, service atomicity, shared schema, and state-root boundaries remain closed on unchanged source. Package, vendor, and harness execution remains blocked or manual without trusted resolver metadata, with no generic vendor shell executor.

Aggressive boundary gate: PASS. 19 verified, 0 unverified, 19 total, 0 deferred, 0 overrides. Focused haocli and vmgateway tests pass, Windows amd64 compilation passes, and all 14 CI checks are successful on this exact SHA.

harshitsinghbhandari (commented):
True verdict: APPROVE. Submitted as COMMENTED because GitHub prohibits approving this account own pull request. Aggressive boundary gate: 19 verified, 0 unverified, 0 deferred. The custom AO_DATA_DIR planner, executor, recovery, daemon unit, and gateway unit agree on the exact path. Pair certificate and passcode identity remain persistent across runtime overrides. Windows managed reads and every mutation primitive use no-reparse held-parent handles. The native windows-latest Go job executed and passed the reparse target, reparse ancestor, ancestor swap, create, mkdir, chmod, remove, and rename tests under race. The previously failing create access, rename handle lifetime, and benign empty read window are corrected. Focused prior security regressions pass locally, Windows amd64 compilation passes, and all 14 required checks, including the full backend race suite, are successful at the guarded head.

PR #157, opened 2026-09-14
feat: add HAO service manager operations

### LINKED ISSUE #155: feat: add internal HAO service manager operations
Implement the internal HAO systemd operation boundary needed by init, lifecycle, and adoption work.

Scope:
* bounded status and journal discovery
* enable, disable, start, stop, and restart for canonical managed units only
* daemon and gateway dependency ordering
* privilege handoff with exit status 3 refusal semantics
* invocation-scoped rollback bookkeeping
* explicit unsupported or manual response outside supported Ubuntu systemd hosts

No new CLI commands are included.

Acceptance:
* daemon starts before gateway and gateway stops before daemon
* healthy matching services are not restarted
* partial activation rolls back only invocation-owned changes
* fixed argv and target unprivileged user invariants are tested
* bounded failures and unsupported platforms are tested

harshitsinghbhandari:
Closes #155

## Summary

* add the internal Ubuntu systemd discovery and operation boundary for canonical HAO services
* enforce daemon-first start and gateway-first stop ordering
* provide bounded status and journal discovery plus fixed privileged mutation argv
* record invocation-owned activation changes and reverse only those changes on rollback
* preserve exit status 3 privilege refusal and explicit manual responses on unsupported hosts

No new CLI commands or root wiring are included.

## Tests

* `GOCACHE=/tmp/hao-services-gocache go test ./internal/haocli -count=1`
* `GOCACHE=/tmp/hao-services-gocache go vet ./internal/haocli`
* `git diff --check`

## Risk

This is an internal boundary only. Real Ubuntu systemd acceptance remains required when later init and lifecycle commands consume it.


backend/internal/haocli/service_manager.go
```diff
@@ -0,0 +1,337 @@
...
+	}
+	if m.journalctl == "" {
+		return "", errors.New("canonical /usr/bin/journalctl is unavailable")
+	}
+	if lines < 1 || lines > maxJournalLines {
+		return "", fmt.Errorf("journal lines must be between 1 and %d", maxJournalLines)
+	}
+	probeCtx, cancel := boundedContext(ctx, m.timeout)
+	defer cancel()
+	output, err := m.observer.Run(probeCtx, m.journalctl, "--no-pager", "--lines", fmt.Sprintf("%d", lines), "--output", "short-iso-precise", "--unit", canonicalServiceUnits[component])
```
harshitsinghbhandari: [P2] Preserve journal output instead of using the diagnostic truncation path. The production systemObserver.Run (observe.go:183-195) truncates every successful result to 256 bytes. Consequently Journal(..., 200) silently returns only the beginning of the requested journal, often a couple of entries, and drops the newest entries needed for diagnosis. Use a dedicated byte-bounded journal reader and cover output longer than 256 bytes through the real reader boundary; the current tiny fake response cannot catch this.

backend/internal/haocli/service_manager.go
```diff
@@ -0,0 +1,337 @@
...
+}
+
+func (m *haoServiceManager) rollback(ctx context.Context, result ServiceOperationResult, failure error) (ServiceOperationResult, error) {
+	for i := len(result.Completed) - 1; i >= 0; i-- {
+		change := result.Completed[i]
+		inverse := map[string]string{"start": "stop", "enable": "disable"}[change.Operation]
+		if inverse == "" {
+			continue
+		}
+		if err := m.mutate(ctx, inverse, change.Component); err != nil {
```
harshitsinghbhandari: [P2] Give rollback a fresh bounded cleanup context. Activate passes its original ctx here. If that context is canceled or its deadline expires after enable/start has succeeded, every inverse mutation inherits the already-canceled context and exec.CommandContext refuses to execute it. The first cleanup fails, leaving all completed activation changes behind. Detach cleanup from caller cancellation while retaining an explicit timeout, and add a context-aware test that cancels after a completed mutation.

backend/internal/haocli/service_manager.go
```diff
@@ -0,0 +1,337 @@
...
+func (m *haoServiceManager) Stop(ctx context.Context, components []string) (ServiceOperationResult, error) {
+	return m.applyOrdered(ctx, components, true, "stop")
+}
+
+func (m *haoServiceManager) Disable(ctx context.Context, components []string) (ServiceOperationResult, error) {
+	return m.applyOrdered(ctx, components, true, "disable")
+}
+
+func (m *haoServiceManager) Restart(ctx context.Context, components []string) (ServiceOperationResult, error) {
+	if _, err := m.Stop(ctx, components); err != nil {
```
harshitsinghbhandari: [P2] Preserve completed stop operations in Restart results. If stopping gateway succeeds but stopping daemon fails, Stop returns the completed gateway stop, but this branch discards it and returns an empty result. If both stops succeed and the subsequent start fails, their records are also lost. Callers therefore cannot recover or accurately report the services this invocation left stopped. Accumulate both phases into the returned ServiceOperationResult, including partial failures, and cover these cases in a Restart test.

harshitsinghbhandari (commented):
## Review: changes requested

Three correctness issues need fixes, detailed inline: journal output is silently truncated, cancellation prevents activation rollback, and Restart drops completed stop records.

### Boundary verification
Static boundary review found that the real observation adapter truncates journal data while the new test substitutes a short fake response. Cancellation-aware cleanup and restart failure bookkeeping are also not covered by the added tests. Real Ubuntu systemd acceptance remains unverified, as noted in the PR description.

Reviewed checkout 40b57853707b45b4c72ffd830c7d65ea49d73fac against origin/develop. Tests and project programs were not executed, and no checkout files were modified.

harshitsinghbhandari (commented):
## Review: approved, ready

Reviewed 80ec292cb2dca1cf03141f19440708f8a04d7a69 against origin/develop and checked the changes since the previous review. All three prior findings are addressed: journal capture now has a dedicated 64 KiB limit, rollback uses fresh per-command bounded contexts, and Restart preserves completed operations from both phases. Added regression tests cover long journal output, cancellation cleanup, and partial stop failure.

No remaining high-confidence correctness findings in this internal-only scope.

### Boundary verification
Static inspection confirms the corrected journal reader and cleanup/result boundaries. The long-output regression uses the actual output adapter. Real Ubuntu systemd acceptance remains outside this review, consistent with the PR scope.

Tests were inspected, not executed. No checkout files were modified.

harshitsinghbhandari (commented):
## Review: approved, ready

Verified checkout and PR head 7cc1f1724c67a4325f38925f754e8d7cde4ec8b0 against develop. The only change since the approved 80ec292 revision is a documentation comment on ServiceOperationResult. Implementation and tests are unchanged; all three earlier findings remain addressed. No new findings or inline comments.

Previous static boundary review still applies. Real Ubuntu systemd acceptance remains required before later lifecycle integration. Tests were not executed, and no checkout files were modified.

PR #158, opened 2026-09-14
feat: add atomic HAO config mutation

### LINKED ISSUE #156: Implement atomic HAO config creation and mutation
## Scope

Implement Task 1 from the HAO completion plan:

- Add interactive and non-interactive schema v1 config creation.
- Add `hao config set <supported-key> <value>` for known schema keys only.
- Preserve the non-secret v1 contract.
- Use per-config locking, safe paths, same-directory temporary files, durable atomic replacement, last-known-good backup, and recovery.
- Cover validation, concurrency, stale and unsafe paths, interrupted writes, backup recovery, and existing reader compatibility.

This issue intentionally excludes init, pair activation, installer, domain, update, and service operations.

harshitsinghbhandari:
## Summary

- add interactive and complete non-interactive schema v1 config creation
- add known-key `hao config set` mutation with valid local and pair transitions
- lock config writes, reject unsafe paths and stale writers, atomically replace durable files, and retain and recover a last-known-good backup
- keep dry-run and rejected edits read-only even when recovery is available
- verify existing config, status, doctor, and setup readers consume generated state

## Tests

- `GOCACHE=/tmp/hosted-ao-gocache go test ./internal/haocli -count=1`
- `GOCACHE=/tmp/hosted-ao-gocache go test -race ./internal/haocli -count=1`
- `GOCACHE=/tmp/hosted-ao-gocache go vet ./internal/haocli`
- `GOCACHE=/tmp/hosted-ao-gocache go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run --path-mode=abs ./internal/haocli/...`
- `GOOS=windows GOARCH=amd64 GOCACHE=/tmp/hosted-ao-gocache go test -c ./internal/haocli -o /tmp/haocli-windows.test.exe`

Closes #156


backend/internal/haocli/config_mutation.go
```diff
@@ -0,0 +1,511 @@
...
+	errUnknownConfigKey   = errors.New("unsupported configuration key")
+	errInvalidConfigValue = errors.New("invalid configuration value")
+	errConfigBusy         = errors.New("another hao config mutation is active")
+	errConfigStale        = errors.New("configuration changed during mutation")
+)
+
+func (s configMutationStore) set(path, key, value string, dryRun bool) (map[string]any, bool, error) {
+	before, err := readValidatedConfig(path)
+	if err != nil {
+		before, err = s.recover(path)
```
harshitsinghbhandari: [P2] Keep recovery read-only until the requested mutation is validated

When the primary configuration is missing or invalid and a valid .bak exists, this call acquires a lock and recoverAndReadConfig immediately calls durableReplace on the primary. Consequently, `hao config set machine.name preview --dry-run` replaces persistent configuration despite promising no writes. An unsupported key or invalid value also restores the primary before returning an error. Read and validate the backup without restoring it during planning, then perform recovery and the edit only in the locked commit path after validation and the dry-run check. Add regressions asserting that both primary and backup remain unchanged for dry-run and rejected edits against a corrupt or missing primary.

harshitsinghbhandari (commented):
## Review

Changes requested. The config-set recovery path writes persistent state before honoring dry-run or validating the requested edit. See the inline finding.

### Boundary verification
Static inspection found a violated CLI-to-storage contract: dry-run can restore the backup onto the primary configuration. Existing recovery tests do not cover dry-run or rejected edits. No tests or project programs were run, per review restrictions.

backend/internal/haocli/config_mutation.go
harshitsinghbhandari: Fixed in b46db036f. Config planning now reads the valid backup as authority without restoring it. The locked commit rechecks both the digest and whether authority came from backup, then combines recovery with the validated mutation. Added regressions proving dry-run, unknown-key, and invalid-value paths leave corrupt or missing primary state and the backup unchanged.

harshitsinghbhandari (commented):
## Review

Changes requested. The previous dry-run and rejected-edit write issue is fixed. One recovery edge case remains: setting a value already present in the backup reports success without restoring a missing or corrupt primary configuration.

### Boundary verification
Static inspection of the develop diff and follow-up fix found the backup-to-primary no-op recovery gap described inline. Existing tests do not cover this case. No tests or project programs were run, per review restrictions.

backend/internal/haocli/config_mutation.go
```diff
@@ -0,0 +1,490 @@
...
+	}
+	data, err := yaml.Marshal(object)
+	if err != nil {
+		return nil, false, err
+	}
+	validated, err := haocontract.ParseConfig(data)
+	if err != nil {
+		return nil, false, fmt.Errorf("%w: %v", errInvalidConfigValue, err)
+	}
+	if bytes.Equal(data, canonicalBefore) || dryRun {
```
harshitsinghbhandari: [P2] Count backup recovery as a change before taking the no-op exit

When readConfigAuthority returns a backup, canonicalBefore describes the backup, not the primary file. If the primary is corrupt or missing and the backup already has machine.name=laptop, `hao config set machine.name laptop` takes this return and reports unchanged with exit 0, leaving the primary unusable. Config show, validate, status, and setup still read the primary, so they continue failing despite the successful set. Treat fromBackup as requiring a commit even when the requested value matches the backup, while keeping dry-run read-only. Add both corrupt-primary and missing-primary regressions for this same-value recovery path.

harshitsinghbhandari: Fixed in 0eaa01a80. Backup authority now counts as a pending change even when the requested value is semantically unchanged. Real mutations restore a valid primary through the locked commit, while dry-run remains read-only. Added corrupt-primary and missing-primary same-value recovery regressions.

harshitsinghbhandari (commented):
## Review

Approved, ready to merge. Reviewed the changes against develop and the follow-up fixes at 0eaa01a8052332abf52f3aa00e92e6eb63feec11. Both previous findings are resolved: planning and rejected edits remain read-only, and same-value mutations restore missing or corrupt primary configuration from the backup. Regression tests cover the reported recovery cases.

### Boundary verification
Static reinspection of planning, backup recovery, locked commit, and reader compatibility found no remaining actionable issues. No tests or project programs were run, per review restrictions.

